PR #494, opened 2026-04-12
refactor(lfs): remove git-lfs binary dependency, document pure-Go architecture

rsnodgrass:
## Summary

- **Delete `internal/gitutil/lfs_repair.go`** + test + adversarial twin test — dead infrastructure that shelled out to `git lfs ls-files` and was gated on `.gitattributes` (which ox never commits), making the repair silently no-op on every real user machine
- **Remove `RepairLFS` field** from `PushOpts` and all 6 call sites across CLI, daemon, and doctor
- **Remove `checkGitLFS` doctor check** that told users to install `git-lfs`
- **Add `.claude/rules/lfs-no-git-lfs-binary.md`** — full architectural rule with rationale, canonical locations, banned patterns, and the GitLab error trap
- **Add `internal/lfs/doc.go`** — package godoc stating the no-binary rule
- **Promote LFS independence to first Key Policy** in AGENTS.md/CLAUDE.md with link to rule file
- **Add `make check-no-git-lfs-shell`** CI gate in Makefile preflight
- **Net: 172 insertions, 979 deletions** — the strongest teaching signal was the code, not the docs

## Why

ox uses the LFS *concept* (pointer files + Batch API) but implements everything in pure Go (`internal/lfs/`). Users almost never have `git-lfs` installed, and `.gitattributes` is intentionally never committed to avoid auto-hydration.

The old repair infrastructure predating `internal/lfs/` was the root cause of the `ox doctor` failure where a ledger with 327 unpushed commits couldn't push due to "LFS objects are missing" — the repair was gated on `.gitattributes`, which never exists, so it silently did nothing.

```mermaid
graph LR
  A[ox doctor --fix] --> B[checkLedgerBranchStatus]
  B --> C[pushLedger]
  C --> D[PushWithRetry RepairLFS=true]
  D --> E[RepairMissingLFSObjects]
  E -->|.gitattributes missing| F["return 0, nil ← BUG"]
  D --> G[git push]
  G -->|LFS objects are missing| H[permanentPatterns: fail fast]
```

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./internal/gitutil/ ./cmd/ox/` — all pass
- [x] `make check-no-git-lfs-shell` — new CI gate passes
- [ ] CI pipeline

## Summary by CodeRabbit

* **New Features**
  * Added an in-repo LFS reconciliation flow to detect and remediate missing LFS pointers before push.

* **Bug Fixes**
  * Improved push retry logic to recover from missing-LFS-object failures via reconciliation.

* **Documentation**
  * New policy and guidance enforcing a pure-Go LFS implementation and disallowing external git-lfs shell-outs; updated test descriptions.

* **Chores**
  * Removed legacy LFS repair utilities/tests and added a preflight check preventing git-lfs shell-out patterns.


rsnodgrass:
All 4 review comments addressed in commit 28df1f83:

1. **ReconcileLFS (bool, error)**: Done. Errors logged internally, not surfaced to users — ledger repair is internal plumbing.
2. **Duplicate OIDs**: Fixed. `oidToIdx` changed from `map[string]int` to `map[string][]int` — all files sharing a missing OID get replaced.
3. **Failed squash = error**: Fixed. Returns error instead of warn+continue.
4. **Makefile grep**: Fixed. Uses POSIX ERE compatible with macOS grep.

PR #607, opened 2026-05-13
fix(codedb): recover Bleve.Batch panics; probe field FSTs in CheckIntegrity

galexy:
## Summary

Diagnosed in session [2026-05-11T23-30-galexy-OxMfxG](https://sageox.ai): a partial-write disk-corrupted Bleve segment left a vellum FST nil. The next `ParseComments` batch panicked in `vellum.(*FST).GetMaxKey`, unwound through `doIndex`'s deferred `db.Close()`, and the bbolt mutex still held by `Batch()` produced a self-deadlock. The daemon wedged for 39 minutes in `sync.Mutex.Lock` while still reporting `indexing_now: true` to every CLI call.

This PR addresses both halves of the failure: it stops the panic from escaping `Batch()`, and it makes `ox doctor` actually capable of detecting the precondition before the daemon hits it.

```mermaid
flowchart LR
    A[ParseComments] --> B["bleve.Batch()"]
    B -->|holds bbolt mutex| C["vellum.GetMaxKey(nil)"]
    C -->|panic| D["doIndex defer db.Close()"]
    D -->|"contends for same mutex"| E[Deadlock — 39 min wedge]
    style E fill:#fdd,stroke:#a00
```

## Changes

**1. `safeBatch` wrapper at every Bleve `Index.Batch` call site** — `internal/codedb/index/indexer.go` + `internal/codedb/index/safebatch.go`
- Recovers panics from `Batch()` and converts them to errors so the caller's deferred `db.Close()` runs normally instead of deadlocking.
- Applied to all 5 sites: code batch flush, diff batch flush, dirty-index batch, comment batch (mid-loop and final).

**2. `CheckIntegrity` probe rewrite** — `internal/codedb/store/store.go`
- Old probe used `bleve.NewMatchNoneQuery()` with `Size=0`, which never touched the FST — the corruption it was supposed to catch was invisible.
- New probe path: `idx.Fields()` → `idx.Advanced()` → `adv.Reader()` → for each field, `reader.TermFieldReader(ctx, probeTerm, field, ...)`. Each `segment.Dictionary(field)` (where `vellum.Load` happens) runs inline on the calling goroutine. Critically, this avoids `FieldDict`, which scorch implements by spawning one goroutine per segment whose panic would escape our `recover`.
- Walks every indexed field, not just `_id`, so corruption on the searchable `content` field is detected.
- Wrapped in `runBleveProbe` which `recover()`s panics and returns `ErrCorrupt`.

### Iteration history

The probe went through four review rounds with Codex before landing here:
1. `MatchAllQuery` — but scorch serves it from `DocIDReaderAll()`, bypassing the FST.
2. `FieldDict.Next()` — but `Close()` was non-deferred, leaking the index read lock on panic.
3. `FieldDict` with `defer Close()` — but scorch's `newIndexSnapshotFieldDict` spawns a goroutine per segment for `Dictionary(field)`; panics in those goroutines escape recover.
4. `Document(id)` — synchronous, but only loads the `_id` FST; misses corruption on `content`.
5. `Advanced() + Reader() + TermFieldReader` per field — synchronous and covers every field. ← shipped

## Scope and known limits

- The probe catches FST loading failures (bad vellum bytes, dict-offset errors).
- It does **not** catch the niche "field present in fieldsMap but `dictStart == 0`, returns a dict with `fst == nil`" pattern — that one only panics during write-path `DocNumbers` calls. `safeBatch` covers it at the actual call site.
- A separate hardening pass (bounded `db.Close()` defer in `doIndex`) is intentionally **not** in this PR — different shape of change, deserves its own review.

## Test plan

- [x] `make lint` clean on touched files
- [x] `go test ./internal/codedb/...` all green
- [x] `safebatch_test.go` — panic, error pass-through, success, non-string runtime panic
- [x] `integrity_test.go` — empty index passes; populated index passes; panic/error/runtime-panic in probe → `ErrCorrupt`
- [ ] Manual: run `ox code index --full` on a real repo, verify no regressions
- [ ] Manual: simulate corruption (truncate a `.zap` segment file mid-FST), confirm `ox doctor` reports `ErrCorrupt`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for database batch operations to prevent crashes and convert panics into recoverable errors.
  * Enhanced database integrity checks with more thorough validation to detect corruption earlier.

* **Tests**
  * Added comprehensive test coverage for error recovery and database integrity validation scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sageox/ox/pull/607)

### DIFF
--- a/internal/codedb/index/indexer.go
+++ b/internal/codedb/index/indexer.go
@@ -132,7 +132,7 @@ func (st *indexState) flushCodeBatch(force bool) error {
 	if !force && st.codeBatchN < bleveBatchSize {
 		return nil
 	}
-	if err := st.store.CodeIndex.Batch(st.codeBatch); err != nil {
+	if err := safeBatch(func() error { return st.store.CodeIndex.Batch(st.codeBatch) }); err != nil {
 		return fmt.Errorf("flush code batch: %w", err)
 	}
 	st.codeBatch = st.store.CodeIndex.NewBatch()
@@ -152,7 +152,7 @@ func (st *indexState) flushDiffBatch(force bool) error {
 	if !force && st.diffBatchN < bleveBatchSize {
 		return nil
 	}
-	if err := st.store.DiffIndex.Batch(st.diffBatch); err != nil {
+	if err := safeBatch(func() error { return st.store.DiffIndex.Batch(st.diffBatch) }); err != nil {
 		st.diffIndexFailed = true
 		slog.Warn("codedb: diff index write failed, type:diff search will be incomplete for this run", "err", err)
 		st.diffBatch = st.store.DiffIndex.NewBatch()
@@ -477,7 +477,7 @@ func BuildDirtyIndex(ctx context.Context, localPath, dirtyPath string, opts Inde
 	}
 
 	if indexed > 0 {
-		if err := dirtyIdx.Batch(batch); err != nil {
+		if err := safeBatch(func() error { return dirtyIdx.Batch(batch) }); err != nil {
 			dirtyIdx.Close()
 			_ = os.RemoveAll(tmpPath)
 			return 0, fmt.Errorf("batch dirty index: %w", err)
@@ -2071,7 +2071,7 @@ sendLoop2:
 				stats.CommentsExtracted++
 
 				if commentBatchN >= bleveBatchSize {
-					if err := s.CommentIndex.Batch(commentBatch); err != nil {
+					if err := safeBatch(func() error { return s.CommentIndex.Batch(commentBatch) }); err != nil {
 						rows.Close()
 						return stats, fmt.Errorf("flush comment batch: %w", err)
 					}
@@ -2102,7 +2102,7 @@ sendLoop2:
 
 	// flush remaining bleve batch after SQL commit
 	if commentBatchN > 0 {
-		if err := s.CommentIndex.Batch(commentBatch); err != nil {
+		if err := safeBatch(func() error { return s.CommentIndex.Batch(commentBatch) }); err != nil {
 			return stats, fmt.Errorf("flush final comment batch: %w", err)
 		}
 	}
--- a/internal/codedb/index/safebatch.go
+++ b/internal/codedb/index/safebatch.go
@@ -1,0 +1,26 @@
+package index
+
+import (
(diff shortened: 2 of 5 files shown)


rupakm:
Folded into **#623** as commit `2e4153e` (authorship preserved — `git log` shows you as author, I'm the committer that did the cherry-pick).

Rationale: PR #623 touches the same two files (`indexer.go`, `store.go`) — separate merges would have conflicted. Folding avoids the rebase dance and keeps both fixes in one coherent "CodeDB robustness" PR.

Two small adaptations during the cherry-pick:

1. **Mid-loop `safeBatch` site removed.** PR #623's CodeRabbit review (point #4) flagged that the mid-loop `commentBatch >= bleveBatchSize` flush could orphan Bleve docs if the SQL tx later rolls back (Bleve commits couldn't be undone, comment IDs they pointed to vanish). So that mid-loop flush block was removed entirely; only the final post-`tx.Commit()` flush remains, and that one IS wrapped in `safeBatch`. Net: 4 of your 5 wrap sites apply unchanged (code, diff, dirty, final-comment); the 5th wrap site went away with the flush block. The deadlock you fixed is still impossible because the panic-capable `Batch` call is the single post-commit flush.

2. **`CheckIntegrity` nil-skip preserved.** Branch A also has a nil-`Index` skip for `OpenSQLOnly` stores. Merged with your `probeBleveIndex` so SQL-only stores don't nil-deref before the recover engages:
   ```go
   for name, idx := range map[string]bleve.Index{...} {
       if idx == nil { continue }
       if err := probeBleveIndex(idx, name); err != nil { return err }
   }
   ```

Verified post-fold: full codedb race suite green (10 pkgs), `golangci-lint` 0 issues, gofmt clean. Closing #607 once #623 merges; happy to keep #607 open for review-attention if that's preferred.

Thanks for the diagnosis + clear write-up — the 39-min wedge root cause + the `FieldDict goroutines escape recover` detail in `probeBleveIndex`'s comment are the kind of evidence that makes the fix obviously correct.

rupakm:
I suggest we close this PR and push #623 (which also contains these changes)

PR #623, opened 2026-05-27
perf(codedb): ADR-018 index-only Bleve + parse perf/memory wins (-76% bleve, -47% steady)

rupakm:
## Summary

CodeDB **memory + on-disk** optimization pass, anchored on **ADR-018** (already merged via #621 — this PR is the implementation). Plus three orthogonal perf wins surfaced by the profiling work.

**Measured on the ox repo (fresh index, all changes active):**

| Axis | Baseline | After | Δ |
|---|---|---|---|
| **Bleve on-disk** | 437 MB | **105 MB** | **−76%** |
| Total codedb on-disk | 534 MB | 177 MB | −67% |
| ParseSymbols peak heap | 1465 MB | 786 MB | −46% |
| ParseComments peak heap | 1127 MB | 520 MB | −54% |
| Steady-state `heapSys` | 1674 MB | 889 MB | −47% |

Plus `debug.FreeOSMemory()` in daemon paths returns the high-water RSS to OS on top (harness doesn't simulate).

## What's in this PR

**ADR-018 implementation — index-only Bleve + snippet re-derivation from source:**
- `f8bbe38` phase 1a — code/comment Bleve mappings flipped to `Store=false` (index-only); per-index `.mapping_version` markers; upgrade nukes+recreates+writes `.needs_reindex_<name>` (piggybacks on existing self-heal flow).
- `5ffd636` phase 1b — committed-code snippet + real `Line` re-derived from git blob via new `Store.ReadBlob` (lazy bare-repo pool, per-repo mutex). Dirty-overlay fragments still come from Bleve (stored).
- `7b85c88` phase 2 (Option D) — diff index also flipped to index-only; empirical measurement on the biggest live codedb showed **81% Bleve saving (672 → 126 MB) on diff alone**.
- `cdea2e7` phase 2 (Option A on top of D) — diff snippet regenerated on read via `Store.DiffSnippet` (LRU-cached, 512 entries); new shared `internal/codedb/diffformat` package keeps the indexer's write format and search's read format byte-stable.
- `aabc2f7` ADR-018 clarification — "preserved vs removed" table + defused the "index-only" jargon trap (Ryan's review concern).

**Orthogonal perf wins surfaced by profiling:**
- `57e1139` chunked `ParseComments` (parseChunkSize=512, bounds the all-results-live set) + daemon `GOGC=50` at startup + `debug.FreeOSMemory()` after each indexing pass.
- `540c81f` search enrichment **N+1 → single batched `WHERE id IN (…)`** per search; also adds the `memprofile` test harness (peak-heap sampling per phase) + the `BleveStoredContentCost` experiment that drove the storage decisions.

## Architecture

```mermaid
flowchart LR
  subgraph Index["Index time"]
    P[parse blob] --> Bleve[(Bleve index-only)]
    P --> SQL[(SQLite metadata)]
    Diff[generate diff text] --> Bleve
  end
  subgraph Read["Search read"]
    Q[query] --> B[Bleve match → doc id]
    B --> E[enrich: batched SQL]
    E --> S{snippet}
    S -->|code| G[git blob → locate match → line + ±ctx]
    S -->|comment| C[SQLite comments.text]
    S -->|diff| D[Store.DiffSnippet: blobs → diffformat.Format → LRU]
  end
```

## Verification

- `go test -race -count=1 ./internal/codedb/...` — all packages green (1232 tests across 10 pkgs).
- `golangci-lint -c .config/golangci.yml ./internal/codedb/...` — **0 issues**.
- `gofmt` clean.
- `memprofile` harness reproduces the headline numbers; `TestBleveStoredContentCost` (74% synthetic) and `TestDiffStoredVsIndexOnly` (81% on 52k real diffs) reproducible on demand.
- 4 mapping-version upgrade-path tests (`mapping_version_test.go`).
- Snippet locator tests + diff-format byte-stability tests.

## Migration / operator impact

- **Reindex required** on existing codedbs to realize the storage savings — handled automatically via the daemon's existing self-heal flow (`.needs_reindex_<name>` markers triggered by `.mapping_version` mismatch). No operator action.
- New search-result semantics: `Line` populated for code hits (was always `0`); `Content` for code/diff is now a plain re-derived snippet, not an ANSI fragment.
- New daemon default `GOGC=50` (only if operator's `GOGC` is unset); cuts steady RSS ~47% at ~10% more GC CPU during heavy phases.

## Open question still on the table
ADR-018 documents Option C@8K as a viable phase-2 follow-up if the diff-snippet regenerate-on-read latency turns out to bother anyone in practice. Not implemented; Option A's empirical numbers made D+A the right shipping default.

## Summary by CodeRabbit

* **New Features**
  * Index schema upgrade flow accepted and applied automatically on startup.
  * Search snippets for code and diffs are re-derived at query time, with an in-process bounded cache.

* **Performance & Optimization**
  * Reduced disk usage by switching primary text fields to index-only storage.
  * Faster search result assembly via batched database enrichment and improved memory/GC handling.

* **Tests & Docs**
  * Added tests, benchmarks, and an ADR documenting the index-only approach and diff/snippet strategy.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sageox/ox/pull/623?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


rupakm:
Addressed @coderabbitai's review — pushed as `a07c67e` on top of `cdea2e7`. Summary of fixes:

| # | File:line | Fix |
|---|---|---|
| 1 | `docs/adr/ADR-018-...md:1` | Added `<!-- doc-audience: ai -->` (ADRs may need AI editing for status/phase updates) |
| 2 | `docs/adr/ADR-018-...md:44,191,200` | Tagged all three previously-unfenced code blocks with `text` |
| 3 | `internal/codedb/index/diff_size_distribution_test.go` | `require.Greater(t, storedSize, 0)` guard before the saving% calculation — empty corpus would emit `Inf/NaN` |
| 4 | `internal/codedb/index/indexer.go` (processCommentChunk) | **Removed the mid-tx Bleve flush at `bleveBatchSize`**. Intermediate flushes were happening before SQL commit → orphan Bleve docs on rollback. Now: accumulate the whole chunk, flush once after `tx.Commit()`. Per-chunk bound (≤ chunk size) keeps the in-memory batch reasonable. |
| 5 | `internal/codedb/search/executor.go` (×3 sites) | Wrapped raw SQL `QueryContext` errors with `fmt.Errorf("query <kind> hit metadata: %w", err)` in `enrichDiffHits` / `enrichCodeHits` / `enrichCommentHits` |
| 6 | `internal/codedb/store/diff_snippet.go` | Honor the docstring's empty-result contract — return `""` (not the header-only `"--- a/path\n+++ b/path\n"`) when both blob reads fail. Cached as `""` too so repeats are free. |
| 7 | `internal/codedb/store/store.go` (bleveMappingFor docstring) | Updated the stale "diff stays on default stored mapping" — now correctly reflects that diff joined the index-only set in phase 2 (with snippet regenerate-on-read via `Store.DiffSnippet`). |
| 8 | `internal/codedb/store/store.go` (`readMappingVersion`) | Distinguish `os.IsNotExist` (legitimate pre-marker → return `(1, nil)`) from other I/O errors (return `(0, err)`). Caller now refuses to upgrade destructively on transient read failures — a flaky filesystem can't trigger a `nuke-and-recreate` of a healthy index. Updated callers + tests. |

**Verification:** full codedb race suite green (10 packages, 0 races), `golangci-lint -c .config/golangci.yml` 0 issues, gofmt clean.

CI re-run was already kicked for the earlier flake in `TestCheckpointFlush_DataAccessibleAfterRun` — passes locally, looks like a test-infra race in repo setup (`git error: invalid object ... Error building trees`), not from this PR.

rupakm:
Folded in @galexy's #607 as commit `2e4153e` — recover Bleve.Batch panics + rewrite `CheckIntegrity` probe to actually exercise field FSTs. galexy preserved as author. See #607 for the rationale + the merge notes on the two small adaptations during cherry-pick.

Stack now: PR scope is **ADR-018 + CodeDB robustness pass**. 9 commits, 1853/-187. Full suite green.

PR #718, opened 2026-07-24
feat(schema): publish raw.jsonl JSON Schema at a stable public $id

rsnodgrass:
## Summary

Anyone can now read a SageOx session recording without importing our Go code — or asking us what a field means. `raw.jsonl` gets a published JSON Schema at a stable public `$id`, MIT-licensed, so other agents and other vendors' tooling can validate, parse, and generate types against our session format directly.

Writing it against real fixtures turned up **seven documented-vs-actual drifts in our own format**, including one PII bug that was about to be published as fact: the spec described `username` as an email and used `dev@example.com` in two examples, while the code has said *"privacy-safe display name — NOT an email"* since it was written. `raw.jsonl` is committed to a ledger repo.

**Approach:** hand-write the line schema (entries have no Go struct — `WriteRaw` takes `map[string]any`), then make it trustworthy with three test legs: it must validate our shipped fixtures, validate live `SessionWriter` output, and reflect `StoreMeta`'s fields so a new Go field can't ship without appearing in the schema.

## What the fixtures taught us

`raw.jsonl` is not one shape. Three dialects occur in practice, and a reader has to handle all of them:

| Dialect | Written by | Header | Entries |
|---|---|---|---|
| Flat | `SessionWriter.WriteRaw()` | `type:"header"` + `metadata` | flat keys, `timestamp`, zero-based `seq`, `eid` |
| Nested | `SessionWriter.WriteEntry()` | same | `{type, timestamp, seq, eid, data:{role, content}}` |
| Import | `capture-prior`, importers | `_meta` wrapper | `ts`, one-based `seq`, `tool`, object-valued `tool_input`, no `eid` |

Seven drifts found, all now documented in the schema's `description` fields where a consumer actually meets them:

| # | Drift | Consequence if unnoticed |
|---|---|---|
| 1 | `username` documented as an email, with two `dev@example.com` examples | PII in a committed ledger repo — **the reason this is worth catching before publishing** |
| 2 | `session_id` means the `ses_` recording identity natively, but an *agent identifier* in the import dialect | a session attributed to the wrong recording |
| 3 | `tool_input`/`tool_output` are string **or** object | a string-only reader silently drops every imported tool call's arguments |
| 4 | `type` vocabulary is 7 values in the wild, 4 in the spec | a reader that enumerates rejects real sessions |
| 5 | `StoreMeta` has `session_id` + `ox_version`; spec documented neither | invisible fields |
| 6 | Aliases: `ts`→`timestamp`, `tool`→`tool_name`, `ox_username`→`username`, `started_at`→`created_at` | half the corpus unparsed |
| 7 | Spec marks `agent_id`/`agent_type`/`timestamp`/`seq`/`eid` required; all are optional in Go or absent in the import dialect | a reader that trusts the table rejects valid sessions |

Nos. 2, 3 and 5 were caught by the tests failing on first run, not by reading — which is the argument for the test legs. No. 7 came from CodeRabbit on this PR; the fix tightens `storeMeta` to require `version`/`created_at` (the only two `StoreMeta` fields without `omitempty`) and corrects the docs table, done now because adding a `required` entry after v1 ships would force a v2 under this schema's own rule.

## Two decisions worth a reviewer's attention

**Open, not closed.** Every object omits `additionalProperties`, and entry `type` is deliberately not an `enum`. A consumer holding a v1 copy of the schema must keep validating files written by a newer ox; `additionalProperties: false` would turn every additive field into a validation break for everyone downstream. Strictness stays on the writer side where we control both ends.

**`pattern` alongside `format`.** Most validators treat `format` as an annotation rather than an assertion by default, so a published schema relying on `format` alone enforces nothing for the third parties it exists to serve. The first test run proved it — `"last tuesday"` validated as a timestamp.

## What changed

| File | Change |
|---|---|
| `schema/v1/raw-jsonl.schema.json` | **new** — the published line schema (`$id`: `https://sageox.ai/schemas/session/v1/raw-jsonl.schema.json`) |
| `schema/README.md` | **new** — usage, the traps above, the open-by-construction rationale, append-only versioning rule |
| `internal/session/rawjsonl_schema_test.go` | **new** — 4 tests / 16 cases |
| `docs/specs/session-raw-jsonl.md` | PII fix, `session_id` dialect caution, missing fields, dialect table, schema pointer |
| `go.mod` / `go.sum` | `santhosh-tekuri/jsonschema/v6` — **test-only**, confirmed absent from `go list -deps` on the production package |
| `THIRD_PARTY_NOTICES.md` | Apache-2.0 entry for the above |

## Test Plan

- [x] `go test ./internal/session/ ./pkg/adapterprotocol/` — both packages pass
- [x] 4 new tests, 18 sub-cases: 3 fixture dialects, live writer round-trip (6 lines: header + 4 entries + footer), `StoreMeta` reflection over 11 fields, 10 rejection cases
- [x] `golangci-lint run ./internal/session/` — **0 findings in the new file** (40 pre-existing errcheck findings in `store_test.go`, untouched)
- [x] `go build ./...` and `go vet ./internal/session/` clean
- [x] Rejection cases prove the schema isn't vacuous: bad `eid` length/charset, negative `seq`, non-RFC3339 timestamp, header without metadata, entry without type, non-`ses_` native `session_id`, `entry_count` as string, empty `metadata`, header missing `created_at`, non-object line

<details><summary>Why hand-written rather than generated</summary>

The monorepo's conversation-schema design generates from Go types, and that's right there — many types, high churn. Here the on-disk entry has **no Go struct at all** (`WriteRaw(data map[string]any)`), so there is nothing to generate from; the prose spec was the only contract. Reflection still guards the one part that *is* typed — `StoreMeta` — which is where a silent field addition would otherwise slip through.

</details>

<details><summary>Deliberately not in this PR</summary>

- **`pkg/adapterprotocol.RawEntry`** — a different contract (what an *adapter hands to ox*, vs what *ox writes to disk*), with its own version constant and its own reference doc. Giving one format two homes is how drift starts.
- **Tool-call pairing and a `reasoning` role on `RawEntry`** — approved separately. Adding `reasoning` is enum growth, so it mints `v2/` under the rule in `schema/README.md`. Publishing v1 first is deliberate: the bump exercises the compat rule before anything depends on it.
- **Converging the three dialects** — the schema documents the split honestly rather than pretending it away. Convergence is real work and wants its own change.
- **Serving the `$id` URL** — the schema is valid and useful as a repo file today; JSON Schema `$id`s are identifiers and need not dereference.

</details>

## Summary by CodeRabbit

* **New Features**
  * Added a machine-readable JSON Schema for validating `raw.jsonl` session recordings.
  * Added support for validating headers, entries, footers, timestamps, ordering, tool payloads, and multiple session dialects.
  * Added documentation covering schema usage, compatibility rules, dialect differences, and versioning.

* **Documentation**
  * Clarified privacy-safe username requirements and session metadata semantics.
  * Documented open-ended entry types and string-or-object tool fields.
  * Added a third-party license notice.

* **Bug Fixes**
  * Added automated validation to detect malformed or non-conforming session records.

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -91,6 +91,7 @@ The following components are licensed under the Apache License, Version 2.0:
 - github.com/opencontainers/go-digest - Copyright (c) Open Container Initiative
 - github.com/opencontainers/image-spec - Copyright (c) Open Container Initiative
 - github.com/pjbgf/sha1cd - Copyright (c) Paulo Gomes
+- github.com/santhosh-tekuri/jsonschema - Copyright (c) Santhosh Kumar Tekuri
 - github.com/spf13/cobra - Copyright (c) Steve Francia
 - go.opentelemetry.io/auto/sdk - Copyright (c) OpenTelemetry Authors
 - go.opentelemetry.io/contrib - Copyright (c) OpenTelemetry Authors
--- a/docs/specs/session-raw-jsonl.md
+++ b/docs/specs/session-raw-jsonl.md
@@ -2,6 +2,19 @@
 
 The `raw.jsonl` file is the primary session recording format produced by `ox session start/stop`. Each line is a valid JSON object. The file captures an agent-to-human conversation session with full provenance for replay, summarization, and auditing.
 
+## Machine-Readable Schema
+
+A published JSON Schema for a single `raw.jsonl` line lives at
+[`schema/v1/raw-jsonl.schema.json`](../../schema/v1/raw-jsonl.schema.json)
+(`$id`: `https://sageox.ai/schemas/session/v1/raw-jsonl.schema.json`, MIT).
+It is validated in CI against this repo's session fixtures **and** against live
+`SessionWriter` output, so it describes what ox actually writes rather than what
+this document wishes were true.
+
+This document stays normative for everything a per-line schema cannot express:
+line ordering, completeness, secret redaction, and the PII boundary on
+`username`. Where the two disagree, that is a bug — please file it.
+
 ## Completeness Requirement
 
 **`raw.jsonl` MUST contain the complete, unfiltered output from the coding agent.** Every conversation turn — user messages, assistant responses, tool calls, tool results, and system messages — must be included verbatim. Do not summarize, truncate, or selectively omit entries.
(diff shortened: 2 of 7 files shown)


rsnodgrass:
Both nitpicks were right and are fixed in f461705d.

**`storeMeta` required fields** — took the suggestion: `version` and `created_at` are now required, since they are exactly the two `StoreMeta` fields without `omitempty` and so are always serialized by the native writer. Worth doing *now* rather than later: under this schema's own append-only rule, adding a `required` entry after v1 ships would force a v2.

**The `agent_id`/`agent_type` half** — you flagged that the docs table marks them required; the docs are wrong, not the schema. Both carry `omitempty` in Go, so they are genuinely optional. Corrected the table rather than tightening the schema. That is a seventh documented-vs-actual drift on top of the six in the description.

**Entry `Required` column** — corrected to `no` for `timestamp`/`seq`/`eid`, with a note that `type` is the only guaranteed entry field.

Added two rejection cases (`metadata:{}` and a header missing `created_at`) so the new constraint is pinned by a test rather than just asserted.

PR #738, opened 2026-07-31
docs(agents): README matrix matches shipped adapters; add Goose prime

rsnodgrass:
## What broke

The README's "Works with your coding agent" table advertised aspiration and hid shipped work.

- It claimed **Cursor records sessions ✅**. There is no Cursor adapter — Cursor gets an instruction-file marker and nothing else.
- It omitted **Gemini, Amp, OpenCode, Droid, Pi, and Aider** entirely — six adapters that build from `Makefile:34` and ship in every release tarball.

The same drift ran downstream. `cmd/ox/init.go` writes a `.sageox/README.md` into **every customer repo** advertising `.cursorrules` and `.windsurfrules` "integrations" we don't have. `docs/guides/agent-compatibility.md` said OpenCode has no SQLite session adapter — it has had one for months — and pointed at a `make compat-matrix` target that doesn't exist.

## The Goose finding

While auditing the agent list, `agentx`'s model of Goose turned out to be stale in every field. Verified against a live Goose install and the upstream repo:

| Surface | What we believed | Reality |
|---|---|---|
| Hooks | `Hooks: false` — "CLI-based" | Full lifecycle hooks since v1.38 (Open Plugins spec) |
| Context files | `.goose/config.yaml`, `.goosehints` | **`AGENTS.md` first**, then `.goosehints`, hierarchically to repo root |
| Sessions | `SupportsSession() == false` | `session_id` on every hook payload; SQLite store |

The context-file finding is the useful one: **Goose already reads the file ox already marks.** Priming Goose costs one registry entry, not an adapter.

Fixed upstream in [sageox/agentx#13](https://github.com/sageox/agentx/pull/13), released as [v0.1.12](https://github.com/sageox/agentx/releases/tag/v0.1.12), bumped here.

## What this PR ships

- **`README.md`** — table rewritten to ship-reality: one row per bundled adapter, plus a marker-only row. The always-✅ "Query past work" column is gone (a column where every cell is identical carries no information), replaced by a line noting `ox query` / `ox murmur` / `ox status` work from any agent on `PATH`. New **"What fires it"** column distinguishes hooks from plugins from instruction-file markers.
- **Orchestrators section** — ox has detected [Block Buzz](https://github.com/block/buzz), Conductor, Gas City, and OpenClaw for a while, and emits a Buzz-specific prime directive, but the README never said so. Now linked and explained.
[...]

Issue #14, opened 2026-02-18
Session recordings misattribute system-injected content as user messages


rsnodgrass:
## Problem

Session recordings in `raw.jsonl` incorrectly classify system-injected content as human ("user") messages. In the sageox.ai session viewer, framework noise appears labeled with the human's name.

Claude Code's JSONL format uses `type: "user"` for ALL messages sent to the model — not just human input. This includes:

- `<system-reminder>` blocks (framework-injected context)
- `<system_instruction>` / `<system-instruction>` blocks
- `tool_result` content blocks (protocol plumbing — already captured from assistant's `tool_use`)
- Skill expansion content (`<!-- ox-hash: ... -->`)
- Plan mode boilerplate ("Entered plan mode.", "Plan mode is active.")
- Context compaction continuations ("This session is being continued from a previous conversation")
- Conductor local command output (`<local-command-stdout>`, `<local-command-caveat>`)

The adapter layer (`ClaudeCodeAdapter.parseLine`) passes all of these through as `role: "user"`, so downstream consumers (web viewer, HTML export, summaries) display them as human messages.

## Expected behavior

`raw.jsonl` should store correct entry types:
- **user** — genuine human messages only (system tags stripped if mixed with real content)
- **system** — framework-injected content
- **skip** — protocol plumbing like `tool_result` blocks (already captured elsewhere)

## Impact

- Session viewer on sageox.ai shows system noise attributed to the human
- AI-generated summaries treat system context as human intent
- Session metrics overcount human participation

PR #739, opened 2026-07-31
feat(goose): session adapter and Open Plugins hook installer

rsnodgrass:
> Stacked on #738 (README/docs truth pass + Goose prime). Review that one first — this branch contains its commit.

## What this ships

Goose becomes a **Silver-tier** AI coworker: sessions record to the Ledger, and hooks fire `ox agent prime` automatically. Before this, Goose got an `AGENTS.md` marker and nothing else.

New `cmd/ox-adapter-goose/` (~900 LOC + ~600 test), plus the ox-side wiring and release plumbing.

```mermaid
flowchart LR
    GS["Goose session starts"] --> HK["hooks.json fires<br/>ox agent hook SessionStart"]
    HK --> PR["ox agent prime<br/>team context in"]
    PR --> REC["find-session, then<br/>drain on every turn"]
    REC --> DB["sessions.db<br/>SQLite read-only"]
    DB --> LG["Ledger"]
```

## Two places this beats the template it was modeled on

`cmd/ox-adapter-opencode` is the only other SQLite reader, so it was the starting point. Goose's schema supports better in two ways worth not copying:

- **Offset is `MAX(messages.id)`, not `COUNT(*)`.** Goose's `id` is `AUTOINCREMENT`, so it is a real watermark. A count-based offset silently replays or skips rows the moment any message is deleted. `TestMaxMessageID_UsesRowidNotCount` demonstrates the divergence directly: after one delete, the count-based watermark replays an entry that already streamed; the rowid watermark replays nothing.
- **Repo correlation uses `sessions.working_dir`,** an actual column, matched exactly then by subdirectory prefix. OpenCode picks "most recent session, no repo filter." `TestResolveSessionID_PrefersWorkingDir` seeds a *newer* session in a different repo to prove recency does not win.

## The hook installer is a new shape for ox

Every existing ox adapter edits one settings file. Goose follows the [Open Plugins](https://open-plugins.com/agent-builders/components/hooks) spec, so hooks live in a plugin **directory**:

```
<scope>/.agents/plugins/sageox/
├── plugin.json        manifest — Goose ignores a plugin without it
└── hooks/hooks.json   event → command rules
```

Design decisions worth reviewing:

- **No shell script on disk.** `command` points straight at `ox agent hook <Event>`. Verified no shim is needed: ox takes the event name from **argv**, and Goose's stdin fields `session_id` / `tool_name` / `tool_input` are exact matches for what `agentx.HookInput` decodes. Unmapped Goose fields survive in `RawBytes`.
- **`OX_PROJECT_ROOT` is baked into project-scope commands.** Goose sends `working_dir` only on *tool* events, so `SessionStart` alone gives ox nothing to locate the repo with and it would walk up from an arbitrary cwd. `config.ResolveProjectRootOverride()` already honors the variable and `findProjectRoot` checks it first. Paths are single-quoted — Goose runs commands via `sh -c`, so a repo path with a space would otherwise split.
- **`matcher` is never emitted.** Goose treats it as a **regex**, and a bare `"*"` is invalid, so Goose logs a warning and **silently skips the whole rule**. Omitting the field is the only unambiguous "match everything." Pinned by `TestInstallHooks_NeverEmitsMatcher`.
- **Uninstall never blind-deletes.** `.agents/plugins/` is a shared namespace. Uninstall strips only rules carrying the `ox agent hook` marker, and removes the directory only when `plugin.json` carries `x-ox-managed`. Two tests cover it: a foreign *rule* survives, and a foreign *manifest* stops the directory delete.
- **`check-hooks` requires the manifest.** A `hooks.json` with no sibling `plugin.json` looks installed on disk but Goose never loads it, so the hooks never fire. Reporting "installed" there would hide a completely dead setup.

## Hook event coverage

Goose fires eleven events. ox installs **seven**:

| Installed | Why |
|---|---|
| `SessionStart` `SessionEnd` `Stop` | prime, finalize, drain |
| `UserPromptSubmit` | whisper push — the primary channel |
| `PreToolUse` `PostToolUse` | incremental drain |
| `PostToolUseFailure` | **not redundant** — Goose fires `PostToolUse` only on *success*, so without this a turn ending in a failed tool call is invisible until the next success or `Stop` |

| Skipped | Why |
|---|---|
| `BeforeReadFile` `BeforeShellExecution` | strict subsets of `PreToolUse` — reading a file and running a shell command are both tool calls |
| `AfterFileEdit` `AfterShellExecution` | strict subsets of `PostToolUse` (and success-only) |

Installing the four would spawn `ox agent hook` **twice per tool call** for zero signal ox does not already have. `TestInstallHooks_EventCoverage` asserts both lists so the reasoning survives contact with the next reader. If you want literal 11-event coverage anyway, it is a four-line change to `hookEvents` — say so and I will flip it.

## agentx

Two releases, both merged and tagged:

- **[v0.1.12](https://github.com/sageox/agentx/releases/tag/v0.1.12)** ([#13](https://github.com/sageox/agentx/pull/13)) — Goose was modeled as hookless and sessionless with the wrong context files. Now implements `LifecycleEventMapper`.
- **[v0.1.13](https://github.com/sageox/agentx/releases/tag/v0.1.13)** ([#14](https://github.com/sageox/agentx/pull/14)) — canonical `HookEventPostToolUseFailure`, mapped for Goose.

So **no `localEventPhases` entry is needed** in `cmd/ox/agent_hook.go` — unlike Gemini, which is still waiting on that migration.

## Two blockers that are easy to miss

- `internal/session/adapters/adapter.go` — `"goose": "generic"` → `"goose": "goose"`. Without this, `lookupAdapter` resolves the alias to the always-registered generic adapter, `GetAdapter`'s `RegisterExternalAdapters()` retry never runs, and the binary is permanently shadowed.
- `internal/session/adapters/session_roots.go` — **fail-closed.** The daemon rejects every `SessionFile` from an adapter with no allow-listed root.

## Drift backfilled while here

The release plumbing was already behind, and Goose would have widened the gap:

| File | Was | Now |
|---|---|---|
| `scripts/install.sh` | 5 of 8 adapters — pi, aider, droid never landed on `PATH` | all 9 |
| `cmd/ox/upgrade.go` | missing droid | complete |
| `internal/adapter/registry.yaml` | **no droid entry at all** | droid + goose |

## Test Plan

- `make lint` — **0 issues**
- `make test` — **17151 tests pass, 1086 skipped, 0 failures**
- `go test ./cmd/ox-adapter-goose/...` — fixture `sessions.db` built in `t.TempDir()` covering all five `content_json` block types, malformed JSON, failed tool envelopes, and the offset semantics

**Verified against a real Goose install on this machine**, not just fixtures:

- `detect` → `{"detected":true,"reason":"found sessions.db"}`
- `find-session` → `goose:20260602_3`, offset `104`
- `read` → 34 entries (1 user, 1 assistant, 32 tool), model `gemma4:26b` extracted from `model_config_json`
- `read-from-offset 100` → 4 entries, `new_offset: 104`; `read-from-offset 104` → empty. **The one-shot path works** — that is exactly the path broken for OpenCode today (`ox-8arr`)
- `install-hooks` → correct plugin directory, quoted paths, no `matcher`, 30s on `SessionStart` and 10s elsewhere; `check-hooks` confirms

**Not yet verified:** whether Goose injects, warns on, or ignores hook **stdout**. `handleStart` pipes `ox agent prime` JSON to stdout. This does not block recording or priming, but it is why I would watch the first real Goose session before assuming `UserPromptSubmit` whispers render cleanly.

## Known limitation, stated not hidden

Goose has **no compaction event**. Team context primed at session start is lost when Goose compacts, and ox cannot work around it. Documented in the compatibility guide, the CHANGELOG, and on `EventPhases()` upstream so nobody "fixes" it by mapping a phase that never fires.

Co-Authored-By: [SageOx](https://github.com/SageOx)

## Summary by CodeRabbit

- **New Features**
  - Added full Goose integration, including session discovery, transcript import, incremental updates, prior-session capture, and repository-aware correlation.
  - Added Goose lifecycle-hook installation, inspection, diagnostics, and safe removal.
  - Goose is now included in supported installation, upgrade, and release workflows.
  - Added read-only `ox recap` evidence reports in text and JSON formats.
  - Added Goose detection and compatibility support across agent setup and diagnostics.

- **Documentation**
  - Expanded compatibility guides with Goose and additional supported AI coding tools.
  - Documented integration methods, session recording, hook behavior, and status reporting.

- **Bug Fixes**
  - Improved OpenCode session reading reliability.


rsnodgrass:
Addressed the review. Four findings were real and are fixed; two I skipped with reasons.

**Fixed**

- **`pluginDir` could resolve relative → `os.RemoveAll` on cwd.** The worst of the four. `pluginDir` now returns `(string, error)`, propagates the `os.UserHomeDir()` failure, and rejects both an empty and a non-absolute root. `check` and `uninstall` previously had no repo-root guard at all — both now do. Covered by `TestPluginDir_RefusesRelativePaths`, `TestHandlers_RejectUnanchoredScopeRoot`, and an end-to-end `TestUninstallHooks_DoesNotTouchCwd` that chdirs into a directory containing a real plugin dir and proves it survives.
- **Unescaped LIKE wildcards in `repoRoot`.** `/repo/my_app` matched `/repo/myXapp/...`, which would attribute another repo's transcript to this Ledger. Now `ESCAPE '\\'` with an `escapeLike` helper. `TestResolveSessionID_EscapesLikeWildcards` seeds a *newer* decoy in the wildcard-matching repo so a leak fails the test.
- **Watermark race.** You were right that the second `MAX(id)` query drops concurrently-inserted rows permanently — Goose writes while ox reads. `readMessages` now returns the highest rowid it actually read, and that is the watermark. The backwards clamp is gone as redundant, as you noted. `TestReadFromOffset_WatermarkComesFromRowsRead` simulates the interleaving directly.
- **`toolResponse` with no payload** now skips instead of emitting an uncorrelatable entry with empty output and no CallID.

**Skipped**

- **`main.go:126` discarded metadata error** — `readMetadata` returns `sql.ErrNoRows` for a session with no model recorded, which is normal, not a failure. Surfacing it would fail imports for ordinary sessions. Metadata is optional in `ImportSessionResult`.
- **`serve.go` error wrapping** — the handlers already return errors that name their operation, and this file matches `ox-adapter-opencode/serve.go` line for line. Diverging one adapter's serve shim is churn without a diagnostic gain.

**Also added, prompted by the review:** `ox doctor` now scans `.agents/plugins/*/{hooks,scripts}` for suspicious hook content. Unlike `.claude/` or `.codex/`, that directory is a **shared namespace** — any tool can install a plugin there and Goose runs its command via `sh -c`. The scan covers every installed plugin, not just ox's own. Plus six `diagnose` tests, including the orphaned-`hooks.json` case where hooks look installed but Goose never loads them.

`make lint` clean, `make test` 17172 pass. Verified against the real binary: unanchored uninstall refuses, and diagnose reports `manifest-missing` on an orphaned hooks.json.

rsnodgrass:
Addressed all open review threads in `7000eaec`. Summary of decisions:

**Fixed:**
- `main.go` — `meta, _ := readMetadata(...)` now logs a warning via `slog.Warn` instead of silently discarding
- `serve.go` — both delegated errors now wrapped with `fmt.Errorf("... %w", err)` for diagnostics
- `session.go` (Goose) — `openDB` DSN now uses `file:` prefix; `created_timestamp` normalized (values > 10 billion treated as milliseconds via `time.UnixMilli`)
- `hooks.go` — (a) `pluginDir` refactored to expose `scopeRoot()` helper; (b) `verifyNoSymlinks()` added and called before any write in `handleInstallHooks`, rejecting repo-controlled symlinks that could redirect writes outside the repo boundary; (c) install now refuses to overwrite a foreign manifest (one without `x-ox-managed: true`), preventing uninstall from later destroying another tool's plugin directory — three new tests cover all three cases
- Docs — "coding agent" → "AI coworker" across four guide files; AGENTS.md reader list updated; Gemini CLI added, OpenAI Agents SDK removed; release note reflects Silver tier; Gemini CLI / Droid now correctly described as "hooks for recording; initial prime via marker" vs. Claude Code / Codex / Goose which truly auto-prime at SessionStart

**Skipped (with rationale):**
- `OX_PROJECT_ROOT` rename → this is an existing internal constant defined in `internal/config/env.go:8` (`EnvProjectRoot = "OX_PROJECT_ROOT"`), consumed by `ResolveProjectRootOverride`. It is not a customer-set variable; ox sets it in the generated hook command so it can locate the repo root without a cwd walk. Renaming it across the codebase changes a published (hook-file-visible) interface and requires human review per project policy — filed for follow-up.

**Bonus fix (OpenCode — ox-8arr):**
- `ox-adapter-opencode` declared `CapIncrementalReader` but never wired `ReadFromOffset` into `adapterruntime.Config` — every hook-driven OpenCode incremental read was silently falling back to full re-read. One-line fix: `ReadFromOffset: handleReadFromOffset` added to the Config struct. Same `file:` DSN prefix applied to OpenCode's `openDB` for consistency.

Issue #710, opened 2026-07-20
Undetected git-lfs misconfiguration silently breaks all session summarization and permanently wedges the ledger


vikgup:
## TL;DR

A host with `git-lfs` **uninstalled but still declared in git config** causes ox to read 130-byte LFS pointer files as if they were session transcripts. Summaries come out empty, validation fails, the daemon retries forever, and the resulting metadata churn produces an unresolvable rebase conflict that permanently blocks ledger sync.

`ox doctor` passed 76+ checks throughout and never once flagged the missing binary.

Five distinct defects below. **#1 and #3 are the high-value fixes.**

**Environment**
- ox `v0.11.1` and `v0.12.0` — reproduced on both; upgrading did **not** fix it
- git-lfs: absent at time of failure; `3.7.1` after fix
- macOS 14 (Darwin 23.6.0), amd64
- Impact: a shared team ledger stopped syncing for all members; self-repair could not converge

---

### Observed state

```
$ which git-lfs
(nothing — not in PATH, no homebrew keg)

$ git config --global --get-regexp 'filter.lfs'
filter.lfs.clean git-lfs clean -- %f
filter.lfs.smudge git-lfs smudge -- %f
filter.lfs.process git-lfs filter-process
filter.lfs.required true
```

git is configured to **require** LFS filters while the binary implementing them does not exist — the residue of an uninstall. Every LFS-backed file then checks out as an unsmudged pointer:

```
$ stat -f%z sessions/<session>/raw.jsonl
129
$ head -1 sessions/<session>/raw.jsonl
version https://git-lfs.github.com/spec/v1
```

### Impact

Every session transcript on the machine was unreadable, yet `ox doctor` reported **76 passed / 1 failed**, where the single failure was the *downstream* rebase symptom rather than the cause. The condition persisted long enough to accumulate 40+ unpushable commits.

### Requested fix

Add a doctor check: if `filter.lfs.*` is configured (especially `filter.lfs.required=true`) but `git-lfs` is not on `PATH`, fail loudly with remediation. This is a two-line check that would have caught the problem immediately.

---

## Defect 2 — Infinite retry when summarizing an LFS pointer

The daemon repeatedly attempted to summarize pointer files, each attempt failing validation:

```json
"summary_status": "failed_validation",
"validation_error": "content validation failed: title too short (0 chars, minimum 3)",
"summary_attempts": 2
```

[...]

PR #742, opened 2026-07-31
fix(session): treat content-store stubs as unreadable, not empty

rsnodgrass:
Part 2 of 4 for #710. **Stacked on #741** — review that first.

## The reported diagnosis is wrong, and the fix depends on saying so

#710 is titled "undetected git-lfs misconfiguration", on the theory that an uninstalled `git-lfs` binary left transcripts as unsmudged pointers.

**ox ledgers ship no `.gitattributes` at all**, so git's LFS filters never applied to them. Those 130-byte files are ox's *own* pointer stubs, written by `lfs.WritePointerFile` — **ledger clones are dehydrated by design**, and content is fetched on demand over the Batch API in pure Go.

The reporter's workaround (install git-lfs, hand-write `.git/info/attributes`, `git lfs fetch`) appeared to work only because ox uploads to the same store. It is also actively dangerous: **by their own account it converts `data/plans/*.md` into pointers on the daemon's next commit**, corrupting the ledger for every teammate.

So Defect 1 as filed is both wrong and would violate `.claude/rules/lfs-no-git-lfs-binary.md`. It's reframed here as a **content-state** check.

## The actual mechanism

```mermaid
flowchart LR
  P["raw.jsonl is a 130-byte pointer"] --> H["HasSubstantiveEntries counts 3 lines, says yes"]
  H --> S["summarizer reads a pointer, emits an empty title"]
  S --> V["validation fails: title too short"]
  V --> R["daily autofix resets the attempt counter"]
  R --> S
  V --> C["duplicate finalize session commits"]
```

`HasSubstantiveEntries` was pure line-counting, and a pointer file has 3 lines. The `MaxSummaryAttempts = 3` cap never bounded anything because a doctor autofix cleared `summary_status` / `summary_attempts` / `validation_error` **daily, before attempting hydration** — re-arming the loop every 24 hours, forever. That's how 21 duplicate `finalize session` commits accumulated across 5 sessions.

## What this ships

- **`ClassifyRawFile` / `RawKind`** — `RawMissing | RawHeaderOnly | RawSubstantive | RawPointerStub`. `HasSubstantiveEntries` is now `== RawSubstantive`.
- **`ResetInlineSummaryEligible` resolves a readable transcript BEFORE clearing terminal state**, and never writes a `.needs-summary` marker pointing at a file that's still a pointer. This is what kills the unbounded loop.
- **`HydrateRawToCacheErr`** — the old function returned `""` on **nine** indistinguishable failure paths, so nothing could report *why* a session wouldn't hydrate or whether retrying was worthwhile. Exactly one sentinel, `ErrNoLFSManifest` (never uploaded → permanently unrecoverable). Everything else — auth, transport, 4xx/5xx, disk — stays retryable. Getting that split wrong in either direction is costly: condemning on a transient 503 throws away a recoverable session; retrying on `ErrNoLFSManifest` recreates the loop. The old signature stays as a wrapper.
- **New doctor check**, `FixLevelSuggested` (not Auto — the fix is network I/O, and the reporter's recovery pulled 159 MB; bounded to 20 sessions/run). Hydration is **cache-only**, per `.claude/rules/cache-only-design.md`.

### The call-site audit — the destructive part

Two callers meant something *different* by `HasSubstantiveEntries` and would have become destructive:

| Call site | If flipped naively |
|---|---|
| `recording.go` ghost cleanup | Deletes `.recording.json` and `removeEmptyDir`s the session directory — **permanent local loss of a synced session** |
| `store.go` `hasRawData` | `ox session list` reports every synced session as empty, since dehydrated is the normal state |

Both now classify explicitly. **I verified the ghost-cleanup guard is load-bearing**: reverting it to the plain call deletes the recording marker of a pointer-backed session.

`session_upload_cmd.go` also gets a pointer-specific message — telling someone their already-uploaded session "has no substantive entries" is actively misleading.

## Test Plan

- **`TestGhostCleanup_PointerStubIsNotAGhost`** — red before, green after; the most consequential test here.
- **`TestGhostCleanup_StillRemovesRealGhosts`** — proves the guard didn't just disable ghost cleanup.
- **`TestResetInlineSummaryEligible_UnhydratablePointerDoesNotReset`** — asserts `meta.json` is byte-identical and no `.needs-summary` is written. This is the proof the loop is dead.
- **`..._HydratedPointerDoesReset`** — the other half, so the fix can't pass by disabling recovery.
- **`TestDehydratedWarning_NeverMentionsGitLFS`** — asserts the user-facing text contains none of `git-lfs`, `git lfs`, `.gitattributes`, `filter=lfs`. Cheapest durable enforcement of the hard rule at the boundary where it actually matters: the text a human reads and acts on.
- **`TestCheckSessionDehydrated_QuietOnHealthyLedger`** — 50 dehydrated-but-summarized sessions must report clean. Dehydration is the *normal* state; flagging it would make doctor noisy enough to ignore.
- Fixed a real ordering bug found by my own test: the nil-client guard fired before the cache check, so an already-hydrated session needed credentials to hand back a file already on disk.
- `make lint` clean; both `check-no-git-lfs-shell` and `check-session-meta-rmw` pass.

## Still to come for #710

Part 3 (stale `rebase-stuck` status never cleared; `rebase --skip` leaking from git's advice block) and Part 4 (skill downgrade guard unreachable; `ox-hash` covers only the body).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,6 +165,28 @@ check-no-git-lfs-shell: ## Ensure no code shells out to git-lfs binary (see .cla
 		exit 1; \
 	fi
 
+check-session-meta-rmw: ## Ensure sessions/*/meta.json is only rewritten via lfs.MutateSessionMeta (ox-q42i, GH #710)
+	@# Rebuilding meta.json from a builder, or reading/marshalling it without
+	@# the flock, silently drops every field the writer doesn't set —
+	@# summary_status, validation_error, summary_attempts, redactions,
+	@# produced_commits, linked_prs, linkage_status. Because the ledger
+	@# auto-resolves sessions/ conflicts to the LOCAL side, that stripping
+	@# then propagates to origin and erases the fields for the whole team.
+	@#
+	@# ALLOWLIST below = the remaining ox-q42i sites, each a first-write or
+	@# an interactive single-writer path. Do not add to it without closing
+	@# that ticket's reasoning; the correct fix is lfs.MutateSessionMeta.
+	@violations=$$(grep -rnE 'lfs\.WriteSessionMeta(Only)?\(' --include='*.go' . 2>/dev/null \
+		| grep -v '_test\.go:' | grep -v 'vendor/' \
+		| grep -vE ':[0-9]+:[[:space:]]*//' \
+		| grep -vE '^\./(cmd/ox/agent_session\.go|cmd/ox/agent_session_recover\.go|cmd/ox/session_regenerate\.go|cmd/ox/session_upload_cmd\.go|cmd/ox/session_repair_meta_summary\.go|cmd/ox/session_push_summary\.go):'); \
+	if [ -n "$$violations" ]; then \
+		echo "$$violations"; \
+		echo "ERROR: rewrite sessions/*/meta.json via lfs.MutateSessionMeta, not WriteSessionMeta[Only]."; \
+		echo "       A fresh-built or unlocked write drops fields it does not set (GH #710, ox-q42i)."; \
+		exit 1; \
+	fi
+
 sync-gitleaks-rules: ## Regenerate the gitleaks-derived detector catalog from a pinned gitleaks.toml
 	@echo "Regenerating internal/session/gitleaks_generated.go..."
 	@cd internal/session/cmd/gitleaks-port && go run . \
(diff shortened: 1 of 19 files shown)


rsnodgrass:
Folded into #741 — reviewing #710 as a two-commit PR rather than a stacked pair. Same content, one review unit.

PR #743, opened 2026-07-31
feat(login): send device_label so users can identify CLI sessions

### LINKED ISSUE #625: ox login: send device_label (hostname/user) so users can identify CLI sessions in web UI
## Problem

`/settings/security` "ox CLI sessions" lists every PAT minted by `ox login`. Users with many machines (laptop, desktop, devcontainer, CI) cannot tell which row maps to which machine — only Created/Last-used/Expires timestamps are shown.

Underlying cause: PAT names are timestamp-only.

```go
// apps/api-go/internal/service/gitlab.go:473  (sageox-monorepo)
tokenName := "sageox-cli-" + time.Now().Format("20060102-150405")
// → "sageox-cli-20260527-143022"
```

DB schema has no `name` column; wire (`cliSessionResponse`) omits one even if added.

## Goal

User can scan the CLI sessions list and immediately identify the machine to revoke.

### ox CLI (this repo)

During `ox login` device-flow token exchange, send a `device_label` field:

```go
// internal/auth/device_flow.go (or login.go)
hostname, _ := os.Hostname()           // e.g. "tacoma"
user, _ := user.Current()              // e.g. "ryan"
label := fmt.Sprintf("%s@%s", user.Username, hostname)
// → "ryan@tacoma"
```

**Privacy:**
- Opt-out via `SAGEOX_NO_DEVICE_LABEL=1`
- Hostname can leak project codename / employer name on shared boxes — document this
- Recommend hostname-only as default if user opt-out feels too coarse; full `user@host` if explicit

### Backend (sageox-monorepo)

- Migration: add `name text` to `gitlab_pats` (nullable for back-compat)
- Accept `device_label` in device-flow token exchange endpoint
- Compose tokenName: `sageox-cli-<label>-<timestamp>` → `sageox-cli-ryan@tacoma-20260527-143022`
- Persist label in `gitlab_pats.name`, mirror to GitLab PAT name
- Add `name string \`json:"name"\`` to `cliSessionResponse`

### Web UI (sageox-monorepo)

`apps/web/src/components/settings/security/CliSessionsList.tsx`:
- Display `session.name` as primary heading
- Fallback to existing "Last used X ago" when `name` is null (covers old rows — back-compat)

## Backward compatibility

Trivially safe. UI today shows no name → old rows continue showing nothing → upgrade adds name only for new logins. No migration needed for existing PATs (display fallback covers).

## Scope split

- **This issue:** CLI sends `device_label` during `ox login`
- **Backend work** (sageox-monorepo): DB migration + endpoint accepts field + response includes `name` + UI renders
- File as separate sageox-monorepo issue once CLI shape agreed

## Out of scope

- Editable labels (user renames on web). MVP = label set at login time, immutable until re-login.
- Device-side rename propagation back to CLI (issue #577 confirmed N/A).

## Origin

Spun out of #577 triage 2026-05-27.

rsnodgrass:
Fixes #625 (CLI half).

`/settings/security` lists every PAT minted by `ox login` with nothing but timestamps. Someone with a laptop, a desktop, a devcontainer and a CI runner cannot tell which row to revoke. This sends a short `user@host` label with the device-flow token exchange.

One edit covers both login entry points — the package-level `Login()` and `(*AuthClient).Login()` share `pollToken`.

## ⚠️ Needs your sign-off before merge

- **New customer-facing env var: `SAGEOX_NO_DEVICE_LABEL`.** Per CLAUDE.md, any new `SAGEOX_*` var needs your review. I took the reversible defaults rather than blocking: **opt-out** (not opt-in), default label `user@host` degrading to host-only.
- **Backend does not accept the field yet** (separate sageox-mono issue, per the issue's own scope split). Standard JSON decoders ignore unknown fields and the CLI reads nothing new off the response, so this is purely additive. **The one real risk: a backend using strict decoding (`DisallowUnknownFields` / Zod `.strict()`) would 400 and break `ox login` outright.** Worth a yes from the mono owner before merge.

## Why `user@host` and not hostname-only

Hostname alone answers the question for a laptop fleet and fails exactly the cases that need it most — shared boxes, CI runners, devcontainers, and multi-account workstations. It's also the format users already read in GitHub SSH key labels, 1Password and `ssh` itself. Degrades gracefully: host-only → user-only → empty.

## Where the username comes from — and three sources deliberately rejected

The OS account name (`$USER`, then `$USERNAME`). Explicitly **not**:

| Rejected | Why |
|---|---|
| `signature.GetMachineID()` | It is the **HMAC key** for the local signature cache — transmitting it leaks a local secret. Also embeds a random UUID that means nothing in a UI. |
| `identity.AttributionName` | Documented **LOCAL ONLY** in that file's own privacy taxonomy. |
| `identity.AttributionUsername` | Resolves through **git config**, so the label would depend on which directory you happened to run `ox login` from — non-deterministic, and more identifying than the account name. |

## Privacy — meeting ADR-008 head-on

A reviewer will reach for "we never collect machine identifiers". That table sits under **Telemetry**, and the distinction is the *scope* of that rule, not an exception to it:

| | Telemetry | Device label |
|---|---|---|
| Audience | SageOx, aggregated across users | The user, on their own account page |
| Identity | Deliberately anonymous | Already authenticated — the server knows who |
| Risk of a machine id | **Correlates an anonymous user's sessions** | None; identity is already established |

A stable machine id is banned in telemetry *precisely because* it de-anonymizes an otherwise anonymous stream. That hazard doesn't exist for metadata attached to a credential the user is knowingly creating. The label answers *which machine*, never *who*.

This PR ships the ADR amendment (new §6 + a summary-matrix row) rather than leaving the scoping to be inferred from section nesting.

**Data minimization applied:** first DNS label only (`laptop`, not `laptop.corp.internal` — the domain identifies an employer more than a machine); allowlisted to `[A-Za-z0-9._-]`; truncated to 32 (user) + 63 (host, the DNS label max, so no legal hostname is ever cut).

## Sanitization is an allowlist, not a denylist

A hostname is attacker-influenceable on a shared box, and the label is transmitted then rendered in a web UI. The allowed set is the intersection of what's safe in a GitLab PAT name, an HTTP header, a JSON string and a URL segment — so everything below `0x20`, `0x7F`, and all non-ASCII is excluded *by construction*:

| Input | Output |
|---|---|
| `a\r\nX-Evil: 1` | `a-X-Evil-1` |
| `"><script>alert(1)</script>` | `script-alert-1-script` |
| `ryan@corp` | `ryan-corp` — an `@` in a *part* is neutralized, so the single `@` in the output is always the composer's |
| 200-char hostname | truncated to 63 |

## Test Plan

- **`TestPollToken_SendsDeviceLabel`** — decodes the actual request body; asserts the pre-existing fields are untouched and the label matches the shape contract.
- **`TestPollToken_OmitsDeviceLabelWhenOptedOut`** — asserts the **key is absent**, not present-and-empty. That's what proves `omitempty`: an empty string would still be a wire change, which an opt-out must not be.
- **`TestDeviceLabel_HostileHostnameStillProducesASafeLabel`** — property test over injection payloads.
- **`TestOptedOutOfDeviceLabel`** — pins truthiness, including that `SAGEOX_NO_DEVICE_LABEL=0` means labels **on**, as it reads.
- `internal/auth/env_naming_test.go` (the `OX_*` guard) still passes; `make lint` clean.

## Follow-up, deliberately out of scope

Not sending the label on the earlier `/api/auth/device/code` request. Showing "you are authorizing ryan@laptop" on the approval screen *before* consent is genuinely better security UX, but it needs a second backend change and an approval-screen design decision.

## Summary by CodeRabbit

* **New Features**
  * Login requests now include a sanitized device label when available, helping identify the associated device.
  * Labels follow safety and length limits and are stored separately from telemetry.

* **Privacy**
  * Set `SAGEOX_NO_DEVICE_LABEL=1` to prevent device labels from being sent during login.
  * Documented device-label handling, transmitted data, storage, and opt-out behavior, including CI/CD guidance.

* **Tests**
  * Added coverage for label formatting, safety limits, hostile input, and opt-out behavior.

### DIFF
--- a/docs/adr/008-privacy.md
+++ b/docs/adr/008-privacy.md
@@ -360,15 +360,27 @@ the distinction is the scope of that rule, not an exception to it:
 | Audience | SageOx, aggregated across users | The user, on their own account page |
 | Purpose | Product analytics | Identifying a credential to revoke |
 | Identity | Deliberately anonymous | Already authenticated — the server knows who |
-| Risk of a machine id | **Correlates an anonymous user's sessions** | None; identity is already established |
+| Risk of a machine id | **Correlates an anonymous user's sessions** | Low; identity is already established, so it adds no linkability |
 
 A stable machine identifier is banned in telemetry precisely because it
 de-anonymizes an otherwise anonymous stream. That hazard does not exist for
 metadata attached to a credential the user is knowingly creating, stored on
-their own account, and displayed back to them. The label answers *which
-machine*, never *who*. Every comparable product (GitHub, Google, 1Password,
-Slack) ships this, because the alternative is a device list of
-indistinguishable rows and no safe way to revoke one.
+their own account, and displayed back to them. Every comparable product
+(GitHub, Google, 1Password, Slack) ships this, because the alternative is a
+device list of indistinguishable rows and no safe way to revoke one.
+
+**Be precise about what it contains.** The label is `user@host`, where `user`
+is the **local OS account name**. So it identifies the machine *and* the
+account on it — which for a single-user laptop is a person. It is not
+anonymous, and this section does not claim otherwise.
+
+What makes that acceptable here, and not in telemetry, is that the request is
+already authenticated: the server knows exactly who is logging in before the
+label arrives. The label therefore reveals nothing about *identity* it did not
+already have; it only adds *which device*. The residual disclosure is the local
+account name itself — e.g. that `r.snodgrass` rather than `ryan` is the account
+on this host — which is why the opt-out exists and why the hostname is reduced
+to its first DNS label.
 
 **Opt-out**, per principle 4:
 
@@ -399,4 +411,4 @@ what ox sent before device labels existed.
 | Guidance sync | None (download only) | `--offline` flag |
 | Prompt delivery | None (download only) | `--offline` flag |
 | Version check | ox version, OS | `ox config set check_updates false` |
-| Device label (`ox login`) | `user@host`, shown only to that user | `SAGEOX_NO_DEVICE_LABEL=1` |
+| Device label (`ox login`) | `user@host` — local OS account + short hostname, shown only to that user | `SAGEOX_NO_DEVICE_LABEL=1` |
--- a/docs/adr/008-privacy.md
+++ b/docs/adr/008-privacy.md
@@ -345,6 +345,51 @@ ox --offline review
 
 ---
 
+### 6. Auth Session Metadata
+
+`ox login` sends a **device label** — `user@host`, e.g. `ryan@laptop` —
+with the device-flow token exchange. It is stored against the PAT and shown
+in `/settings/security` so a user can tell their laptop from their
+devcontainer from their CI runner when deciding which session to revoke.
+
+**This is deliberately outside the "What We NEVER Collect" rule above**, and
+the distinction is the scope of that rule, not an exception to it:
+
+| | Telemetry (§1) | Device label |
+|---|---|---|
+| Audience | SageOx, aggregated across users | The user, on their own account page |
+| Purpose | Product analytics | Identifying a credential to revoke |
+| Identity | Deliberately anonymous | Already authenticated — the server knows who |
+| Risk of a machine id | **Correlates an anonymous user's sessions** | None; identity is already established |
+
+A stable machine identifier is banned in telemetry precisely because it
+de-anonymizes an otherwise anonymous stream. That hazard does not exist for
+metadata attached to a credential the user is knowingly creating, stored on
+their own account, and displayed back to them. The label answers *which
+machine*, never *who*. Every comparable product (GitHub, Google, 1Password,
+Slack) ships this, because the alternative is a device list of
+indistinguishable rows and no safe way to revoke one.
+
+**Opt-out**, per principle 4:
+
+```bash
+SAGEOX_NO_DEVICE_LABEL=1 ox login
+```
+
+The field is `omitempty`, so opting out makes the request byte-identical to
+what ox sent before device labels existed.
+
+**Data minimization applied:**
+- Only the first DNS label of the hostname (`laptop`, not
+  `laptop.corp.internal`) — the domain identifies an employer far more than
+  it identifies a machine.
+- The OS account name, not a git-config or profile name.
+- Allowlisted to `[A-Za-z0-9._-]`, truncated to 32 (user) + 63 (host).
+- Never the local machine id from `internal/signature` — that value is an
+  HMAC key, and transmitting it would leak a local secret.
+
+---
+
 ## Privacy Summary Matrix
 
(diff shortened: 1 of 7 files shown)


rsnodgrass:
**Fixed** — you were right, and it mattered.

`user@host` includes the OS account name, so "never *who*" and "risk: None" were both overstatements in a privacy doc, which is the worst place to overstate.

ADR-008 §6 now says plainly that the label identifies the machine **and the account on it** — for a single-user laptop, a person — and that it is not anonymous. The argument is narrowed to the thing that's actually true: the device-flow exchange is already authenticated, so the label adds *which device* without adding linkability the server didn't already have. The residual disclosure (the local account name itself) is named explicitly as the reason the opt-out and the first-DNS-label reduction exist. Same correction applied to the `deviceLabel()` doc comment and the summary-matrix row.

rsnodgrass:
## Backend check: safe to merge, but **inert** until sageox-mono changes

Verified against `sageox-mono` (`apps/web/app/api/v1/device/token/route.ts`):

**No strict decoding — zero break risk.** The handler does `const body = await request.json()` and destructures `{ device_code }`. No zod, no `DisallowUnknownFields`, no schema validation. An unknown `device_label` is silently ignored. The risk I flagged in the original description does not exist.

**But the field goes nowhere.** The route doesn't forward the client's body — it constructs its own internal request to Better Auth with an explicit `{device_code, grant_type, client_id}`. So `device_label` is **dropped at the first hop**.

So this PR is groundwork, not a working feature. It's safe to land, but nothing changes in `/settings/security` until mono is updated to accept the field, persist it on `gitlab_pats.name`, return it in `cliSessionResponse`, and render it. (PAT naming is still timestamp-only at `apps/api-go/internal/service/gitlab.go:474`, so the issue's premise does hold.) Tracked as `ox-11pk`.

## Second thought on `SAGEOX_NO_DEVICE_LABEL` — the shape is wrong

I shipped the env var because #625 specified it, but ox's established precedent for a privacy toggle is **command + user-config + env var** — telemetry has `ox telemetry off`, `UserConfig.TelemetryEnabled`, *and* `SAGEOX_TELEMETRY`. An env var alone is the weakest third: undiscoverable, not persistent across shells, invisible in `ox config`, easy to forget in CI.

Worth weighing against that: ox already sends far more identifying data to this same authenticated backend — git remote URLs (org names), repo names, session transcripts containing file paths and code. A short hostname plus OS account name isn't incrementally identifying against that baseline, and **nothing else ox sends has a per-field opt-out**.

Two coherent options; the current half-precedent is the one that isn't:

1. **Drop the env var** — no opt-out, consistent with every other field ox transmits.
2. **Promote to `UserConfig`**, env var stays as the CI override — consistent with telemetry.

Since the label is inert today there's no urgency either way. Flagged as `ox-kttu` for your call; happy to implement whichever.

Issue #217, opened 2026-03-14
bug: no filesystem lock on team context directory — concurrent clone/pull races across daemon instances


rsnodgrass:
## Background

When two separate git clones of the same project are open simultaneously, each spawns its own daemon process. Each daemon has its own in-memory `cloneSem` semaphore and `cloneInFlight` sync.Map, so there is **no cross-process coordination** when both daemons operate on the same team context directory on disk.

## Problem

Two concurrent daemon instances can race on the same team context git directory:

- **Concurrent clone**: Both daemons call `git clone` into the same target path. The second one gets `destination path already exists` and triggers unnecessary exponential backoff, burning time and log noise.
- **Concurrent pull**: Both daemons run `git fetch` + `git pull --rebase` on the same repo simultaneously, causing `.git/index.lock` contention or a corrupted rebase state if one process holds the lock while the other tries to write.

This is reliably reproducible in multi-clone E2E test scenarios where two schedulers are pointed at the same team context URL.

## Expected Behavior

Only one daemon at a time performs clone or pull on a given team context directory. Other daemons wait (or skip with a logged warning) rather than racing.

## Proposed Fix

Add a **filesystem lock file** (e.g., `<repo-path>/.sageox/.sync-lock`) acquired exclusively during clone and pull operations. Use `syscall.Flock` (POSIX `flock(2)`) or an equivalent cross-platform advisory lock library. This lock file approach works across daemon processes without any shared memory.

A shared (read) lock during pull and an exclusive lock during clone would also allow safe future parallelism between reads.

## Affected Code

- `internal/daemon/sync.go` lines 1651–1844 — `Checkout()` / clone path
- `internal/daemon/sync.go` lines 2131–2261 — `cloneInBackground()`
- `internal/daemon/sync.go` lines 2280–2370 — `pullTeamContext()`

## Acceptance Criteria

- [ ] Filesystem lock is acquired before clone begins and released after it completes (success or failure)
- [ ] Filesystem lock is acquired before pull begins and released after it completes
- [ ] A second daemon waiting on the lock logs a debug/info message rather than erroring
- [ ] Multi-clone E2E test (`internal/daemon/sync_multiclone_test.go`) passes without races under `-race`
- [ ] No regression in single-daemon behavior

rsnodgrass:
## Design guidance from Ryan

**Avoid `flock()` as the default solution.** Held locks are fragile — a crashed/hung process can hold a lock indefinitely, blocking all other daemons.

[...]

PR #811, opened 2026-08-24
fix(ledger): refuse to auto-commit unresolved conflict markers

### LINKED ISSUE #749: bug: failed rebase autostash commits conflict markers into session meta.json, corrupting the ledger (#601 signature still reproducing on 0.12.0)
## Summary

The `#601` failure signature still reproduces on **v0.12.0** (latest release), and it now has a worse consequence than the one that issue describes: the failed **rebase autostash pop** writes `<<<<<<< Updated upstream` / `>>>>>>> Stashed changes` markers into session `meta.json` files, and ox **commits them**. The result is unparseable JSON permanently in the ledger.

I found **8 corrupted `meta.json` files** in one repo's ledger, spanning **2026-05-07 → 2026-07-30**. Six of them post-date #601's close (2026-05-27).

This is not just a sync failure — it is silent data corruption of the records `ox query` indexes.

## Evidence: same root cause as #601

Every one of the 8 carries the exact signature from #601:

```json
"summary_status": "unrecoverable",
"validation_error": "content validation failed: title too short (0 chars, minimum 3)",
<<<<<<< Updated upstream
  "summary_attempts": 3,
=======
  "summary_attempts": 2,
>>>>>>> Stashed changes
```

Local sits at `failed_validation`, `attempts: 2`; upstream is terminal at `unrecoverable`, `attempts: 3` — the same divergence #601 documented (it saw `attempts: 1` vs `3`).

## The new defect: the autostash, not the rebase

#601 covered the finalize/rebase loop. What it did not cover is the **autostash**.

`ox`'s repair path runs `git pull --rebase`, which creates an autostash. Observed in the failure output:

```
git pull --rebase failed during retry: Created autostash: 7554c4a888
Auto-merging sessions/<redacted>/meta.json
CONFLICT (content): Merge conflict in sessions/<redacted>/meta.json
error: could not apply 812dbf02... finalize session <redacted>
```

When the rebase aborts, the autostash pop conflicts. The markers it leaves say `Updated upstream` / `Stashed changes` — **stash-pop markers, not merge markers** — which is how these are distinguishable from ordinary merge debris. ox then commits the file with the markers still in it.

Nothing validates the file before committing, so corruption is durable and silent.

## Why it can never self-heal

`pull --rebase` replays **every** local-only commit through the same conflict. This ledger was at **540 ahead / 142 behind**. Each retry re-hits the first conflicting commit and aborts, so the backlog only grows. Pushes had been failing continuously for **5 days** before anyone noticed — sessions kept recording locally (43 of the last 50 were `status: local`, exactly 1 `uploaded`), so from the user's side everything looked normal while nothing reached the team.

A merge is the right strategy for an append-mostly ledger. Dry-run confirmed the entire 540-vs-142 divergence reduced to **3 conflicts**, resolvable in a single pass:

```
git merge-tree --write-tree HEAD origin/main
```

## Impact

`meta.json` with conflict markers is not valid JSON. Any consumer that parses session records — the query indexer being the obvious one — silently loses those sessions. In this ledger `ox query "<anything>"` returned `{"team_context":{"results":[]}}` for every probe while the corruption was present.

## Suggested fixes

1. **Never commit a file containing conflict markers.** A JSON-parse (or marker-grep) check before `git add` in the repair path would have caught all 8. This is the highest-value guard — it converts silent corruption into a loud, actionable failure.
2. **Use merge, not rebase, to reconcile the ledger.** Append-mostly history with per-session directories merges cleanly; rebase multiplies one conflict by the number of pending commits.
3. **Do not `--autostash` a repo the daemon is concurrently writing.** A failed pop is what produces the corruption. Either take a real lock for the duration, or refuse to reconcile while the daemon holds writes.
4. **Add an `ox doctor` check** that greps tracked `meta.json` for conflict markers / parse failures. It would have surfaced this in May instead of August.
5. Finish #601's guard: if upstream says a session is `unrecoverable`, the local finalizer must stop re-entering it from `failed_validation`.

## Workaround (what actually repaired it)

`git reset --hard origin/main` (the #601 workaround) discards local-only sessions, which was not acceptable here — 540 commits of real work. Instead:

```bash
cd <ledger>
git branch backup/pre-repair HEAD          # safety net
ox daemon stop                             # stop concurrent writes
git merge origin/main                      # 3 conflicts, not 540 replays
# resolve the 3 by hand, then strip markers from the 8 corrupted files
git add sessions/ && git commit && git push
```

Resolution rule for the conflicts: keep the side with a real `summary`; where neither has one, keep the higher `summary_attempts` (the terminal state). After repair, all **687** `meta.json` files in the ledger parse.

One caution for anyone scripting the cleanup: a naive `grep -rl '<<<<<<<' sessions/` produces false positives. A recorded `session.md` legitimately contained conflict markers because the session itself was about resolving a merge conflict. Only `meta.json` should be auto-repaired.

## Environment

- `ox 0.12.0` (latest release), `go1.26.0`, `darwin/amd64`
- Claude Code adapter, daemon running via `ox daemon start --foreground`
- Ledger: 540 ahead / 142 behind at time of repair; 687 sessions; LFS pointer-only (no local objects)
- Corrupted files dated 2026-05-07, 05-18, 05-22, 07-17, 07-18, 07-19, 07-29, 07-30

Related: #601 (same root cause, closed), #444 (re-finalization overwriting good summaries), #710 (wedged ledger).

aditeya08varma:
## What broke

`ox doctor`'s auto-commit (`fixLedgerDirtyWorkdir`) runs `git add -A` then commits whatever's staged. `git add` does not refuse a conflicted path — it stages the file's current content, conflict markers included, as "resolved." A `git stash pop` conflict from the daemon's `pull --rebase --autostash` leaves no `MERGE_HEAD`/`rebase-merge` marker (a stash pop isn't a resumable git operation), so `detectInProgressGitOp` can't see it either — it just sits there until an unrelated dirty file trips the auto-commit, and `-A` sweeps the conflicted file in too. The markers get baked into the ledger's history permanently.

## What this PR ships

- `internal/gitutil/conflict.go` — `HasConflictMarkers(path)`, a shared guard
- `fixLedgerDirtyWorkdir` (`cmd/ox/doctor_ledger_git.go`) now re-checks for unmerged (U-state) paths immediately before staging and refuses to commit if any are found, instead of blindly running `git add -A`
- The same guard is applied to `commitAndPushLedger` and `commitPointerRewriteAndPush` (`cmd/ox/session_upload.go`) — the routine session-stop commit paths — so the same class of bug can't land through a second door

## Test plan

- `TestFixLedgerDirtyWorkdir_RefusesUnrelatedDirtyFileWithConflictPresent` reproduces the exact shape from this issue: a real `git stash pop` conflict sitting alongside a genuinely dirty, unrelated file, in a real git repo. Verified it fails without the fix (produces the literal "ox doctor: auto-commit ledger changes" commit containing the markers) and passes with it.
- `TestHasConflictMarkers` / `TestFirstConflictMarkerFile` cover the guard in isolation.
- `go build ./...`, `go vet`, `gofmt` all clean.
- Full `internal/gitutil` and relevant `cmd/ox` suites pass.

Closes #749.

---
<a href="https://app.blacksmith.sh/sageox/codesmith/ox/pr/811"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790129528&installation_model_id=433550&pr_number=811&repository=sageox%2Fox&return_to=https%3A%2F%2Fgithub.com%2Fsageox%2Fox%2Fpull%2F811&signature=e8a5be9f17ddd43bcf1fa03b8ce6e695ecb2be643099b597fd242bb06cd94734"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented automatic commits when unresolved Git conflicts or conflict markers are detected.
  - Blocked uploads containing conflicted staged content, including unrelated files.
  - Improved handling of staged renames, copies, deletions, and unusual file paths.
  - Added actionable manual-resolution guidance for conflicts without an abortable operation.
  - Preserved conflicted files and markers during stash-pop conflict scenarios.

- **Tests**
  - Expanded coverage for conflict detection, staged changes, renames, missing files, and upload prevention.

### DIFF
--- a/cmd/ox/doctor_ledger_git.go
+++ b/cmd/ox/doctor_ledger_git.go
@@ -992,6 +992,33 @@ func fixLedgerDirtyWorkdir(ledgerPath string, fileCount int) checkResult {
 			fmt.Sprintf("git add error: %s", strings.TrimSpace(string(output))))
 	}
 
+	// The pre-add U-state check above only catches a LIVE conflict — one
+	// `git add -A` hasn't touched yet. But `git add` on a conflicted path
+	// doesn't just stage it, it resolves the U-state (git now considers it
+	// "modified", not "unmerged"), while the staged content is still the
+	// literal markers. A file that was already staged-with-markers before
+	// this function ran (never U-state at all — e.g. plain-modified, or
+	// unrelated leftover content from an interrupted prior operation) is
+	// invisible to the pre-add check for the same reason. firstUnstageableFileInIndex
+	// (session_upload.go) covers exactly this by reading the actual staged
+	// blobs post-add, the same guard the session-stop commit paths use —
+	// parity here closes the gap instead of leaving doctor's auto-commit
+	// weaker than the routine commit path it's meant to be backing up.
+	if conflicted, err := firstUnstageableFileInIndex(ledgerPath); err != nil {
+		return FailedCheck("Ledger clean workdir",
+			"unable to verify staged content is conflict-free",
+			fmt.Sprintf("checking staged blobs: %s", err))
+	} else if conflicted != "" {
+		return FailedCheck("Ledger clean workdir",
+			"unresolved conflict staged, refusing to auto-commit",
+			fmt.Sprintf("%s still carries unresolved conflict markers even though `git add` "+
+				"accepted it. Committing would bake the markers into the ledger permanently. "+
+				"Resolve it manually:\n       cd %s\n       git status\n       "+
+				"git checkout --ours %s   # or --theirs, depending on intent\n       "+
+				"git add %s && git commit",
+				conflicted, ledgerPath, conflicted, conflicted))
+	}
+
 	// commit via RunGit: it owns the commit.gpgsign=false override plus the
 	// GIT_TERMINAL_PROMPT=0 / cmd.Dir safeguards, so the auto-commit can't
 	// drift from the managed-git execution contract.
--- a/cmd/ox/doctor_ledger_git.go
+++ b/cmd/ox/doctor_ledger_git.go
@@ -930,6 +930,40 @@ func checkLedgerCleanWorkdir(fix bool) checkResult {
 
 // fixLedgerDirtyWorkdir stages and commits all changes in the ledger.
 func fixLedgerDirtyWorkdir(ledgerPath string, fileCount int) checkResult {
+	// Re-check for unmerged (U-state) paths immediately before staging.
+	// checkLedgerCleanWorkdir deliberately excludes U-state entries from the
+	// dirty count that triggers this fix — a lone conflict is
+	// checkLedgerUnmergedPaths's job, not this one's. But that exclusion is
+	// COUNTING logic only; it does not reach the `git add -A` below. `-A`
+	// stages every changed path unconditionally, including any conflicted
+	// one sitting alongside genuinely dirty files that DID trigger this fix.
+	// git has no concept of "refuse to stage a conflict" — `git add` on a
+	// conflicted path takes its current working-tree content, markers
+	// included, and marks it resolved. The commit that follows would then
+	// bake those markers permanently into the ledger. This is the exact
+	// mechanism behind #749: a `git stash pop` conflict (from the daemon's
+	// `pull --rebase --autostash`) leaves UU files with no MERGE_HEAD/
+	// rebase-merge marker — autostash pop isn't a resumable operation, so
+	// detectInProgressGitOp can't see it either. They just sit there until
+	// an unrelated dirty file triggers this auto-commit and sweeps them in.
+	// Refuse instead of guessing at a resolution.
+	statusCmd := exec.Command("git", "-C", ledgerPath, "status", "--porcelain=v1")
+	if statusOut, statusErr := statusCmd.Output(); statusErr == nil {
+		if unmerged := parseUnmergedPaths(string(statusOut)); len(unmerged) > 0 {
+			sample := unmerged[0].Path
+			if len(unmerged) > 1 {
+				sample = fmt.Sprintf("%s (+%d more)", sample, len(unmerged)-1)
+			}
+			return FailedCheck("Ledger clean workdir",
+				"unresolved conflicts present, refusing to auto-commit",
+				fmt.Sprintf("%d unmerged file(s) at %s, e.g. %s.\n       "+
+					"Run `ox doctor --fix` again — the ledger unmerged-paths check "+
+					"resolves these first; auto-committing over them would bake the "+
+					"conflict markers into the ledger permanently.",
+					len(unmerged), ledgerPath, sample))
+		}
+	}
+
 	// ensure .gitignore exists and untrack any cache files BEFORE staging.
 	// without this, git add -A will commit local-only files like sync-state.json.
 	gitserver.EnsureGitignoreBeforeCommit(ledgerPath)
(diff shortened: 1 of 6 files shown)


rsnodgrass (commented):
Thanks for this — it's a careful, well-argued fix, and the test suite is exemplary: red-first, real-git integration reproducing the exact #749 autostash-pop shape, with sanity checks that each setup produced the intended git state. The prior CodeRabbit/Greptile rounds (staged-blob-vs-working-tree, `-z` rename records, modify/delete) are all addressed. Three notes before merge:

**1. The open Greptile P1 (colon-prefixed staged paths) is valid.** In `firstUnstageableFileInIndex`, `git show ":"+rel` misparses a staged path literally named like `1:foo` — git reads `:1:foo` as *stage-1 of `foo`*, not the literal path — so the blob read errors and the guard fail-closed blocks the whole session commit/push. Unlikely for real ledger paths, but it's a latent self-block on the very path the guard is meant to protect. Disambiguate the pathspec (`":0:"+rel` or `":./"+rel`) and add a colon-prefixed-name regression case.

**2. The doctor path is a weaker guard than the session-upload path.** `fixLedgerDirtyWorkdir` (`doctor_ledger_git.go`) only checks `git status` for U-state paths *before* `git add -A`; it never runs `firstUnstageableFileInIndex`. So a file carrying marker text that is plain-*modified* (or already staged) rather than `UU` — the class your own `TestFirstUnstageableFileInIndex_CatchesUnrelatedAlreadyStagedConflict` protects session-upload against — can still be swept into a doctor auto-commit. Both paths write the same ledger; consider routing `fixLedgerDirtyWorkdir` through `firstUnstageableFileInIndex` after staging, for parity.

**3. Heads-up on CI (not your change):** the failing `test` job is `TestDI_PushMurmurCommitsUsesInjectedGitRunner` in `internal/daemon` — a murmur test unrelated to these ledger changes, almost certainly flaky or a main-branch issue. A re-run should clear it; nothing to fix here.

Core change is sound; happy to merge once #1 is addressed. Thanks again for the thorough work.

aditeya08varma:
Real observation, but I don't think it's a new problem. It's an existing, already-accepted tradeoff in this codebase showing up in a new place.

The daemon and CLI already write to the same local git index with no cross-process coordination. That's documented and deliberate, not an oversight. `.claude/rules/daemon-git.md` states the daemon's `ledgerMu` "cannot serialize against the CLI, which runs as a separate OS process," and that push retry is "best-effort, not transactional." The team has already chosen to handle concurrent-writer races with retry instead of prevention everywhere else on this path (see `pushLedger`'s `pull --rebase` retry loop for the remote side). This guard's check-then-commit window is the same class of race, just local instead of remote. It's not a gap this PR introduces, and it doesn't make anything worse than the status quo.

Given the team's own precedent here, and Ryan's explicit guidance on the related #217 to avoid `flock()` as the default and prefer lock-free approaches, I don't think this needs new tracking. It's the same accepted tradeoff as every other ledger writer, not a new discovery. Flagging the reasoning here in case anyone disagrees and wants it filed separately. Happy to open an issue if so.

Issue #854, opened 2026-09-02
ox init: a failed init deletes agent hook files it did not create


deepkawal:
When `ox init` fails after the integration step, its rollback deletes agent hook files
outright — including ones that existed before the run, and including the user's own hook
entries merged into them. Rollback is supposed to restore what was there; here it removes it.

Checked against `main` @ `19fac267` (v0.14.3 + 13).

---

## What happened

A repo already initialized for SageOx, with a working `.codex/hooks.json`. Re-running
`ox init` with a bad `--team` value fails at registration, and rollback runs:

```
ROLLING BACK
✓ Restored .gitignore
✓ Restored .gitattributes
✓ Restored AGENTS.md
✓ Restored CLAUDE.md
✓ Removed hooks.json          ← deleted, not restored
Error: failed to register with SageOx API: HTTP 400 from
https://sageox.ai/api/v1/repo/init: {"success":false,"error":"team name not found: ..."}
```

Four files were correctly *restored*. `.codex/hooks.json` was **removed**. It pre-existed
the run. The file was recoverable only because a copy happened to still be in the git index;
had it not been staged, the content would be gone.

Note what a bad CLI argument cost here: the command rejected the input server-side and then
deleted a config file as part of cleaning up.

## Why

`cmd/ox/init.go:730-734` tracks every path an adapter reports, unconditionally, as a file
**ox created**:

```go
installedHooks := installAgentHooks(gitRoot, true, selectedAgents)
for _, hookFile := range installedHooks {
	tracker.trackCreatedFile(hookFile)
	tracker.trackForceStage(hookFile)
}
```

The tracker's two modes are not interchangeable — `init.go:1711` and `:1728`:

| Method | Rollback behaviour |
|---|---|
| `trackCreatedFile` | `os.Remove(file)` — deletes it (`init.go:1841-1850`) |
| `trackModifiedFile` | restores the snapshotted original (`init.go:1830-1838`) |

So any pre-existing file that reaches `trackCreatedFile` is deleted rather than restored.

**Adapters report merges as writes.** For external adapters, `installAgentHooks` appends
`result.FilesWritten` whenever `result.Installed` is true (`init.go:2104-2115`), and the
codex adapter sets both unconditionally after a **merge** into the existing file
(`cmd/ox-adapter-codex/hooks.go:57-86`):

```go
hooksMap, rawMap, err := readHooksFile(hooksPath)   // reads what is already there
for _, event := range codexHookEvents {
[...]

PR #856, opened 2026-09-02
fix(init): resolve --team by name, slug, or ID before registering

### LINKED ISSUE #855: ox init --team takes three different vocabularies and documents one
`--team` is documented as taking a team ID. It also accepts a team **name**, case-sensitively.
It rejects the **slug** that `ox team list` prints and that every other `ox team` subcommand
accepts. Which of the three you supply is not knowable from the help text, and guessing wrong
costs more than an error message.

Checked against `main` @ `19fac267` (v0.14.3 + 13). Team names below are redacted; everything
else is verbatim.

---

## What happens

```
$ ox init --team platform
Error: failed to register with SageOx API: HTTP 400 from
https://sageox.ai/api/v1/repo/init: {"success":false,"error":"team name not found: \"platform\""}

$ ox init --team Platform
✓ Registered with SageOx (team: team_abc123)
```

Same team. The lowercase form is what `ox team list` displays in its Slug column, and it is
what `ox team show platform` accepts.

The flag's help text (`cmd/ox/init.go:141`):

```go
initCmd.Flags().StringVar(&initTeamFlag, "team", "", "team ID to associate this repo with")
```

So the documented vocabulary is the ID; the accepted vocabularies are the ID and the
exact-case name; the rejected one is the slug the CLI itself prints.

The server's own error says `team name not found`, confirming the value is being matched
against names — the flag is not ID-only in practice, only in documentation.

## Why it happens

`--team` short-circuits team resolution entirely (`init.go:412-415`):

```go
if initTeamFlag != "" {
	// use explicitly provided team
	selectedTeamID = initTeamFlag
} else {
	// fetch teams from API to determine if selection is needed
```

The flag's raw string is assigned straight to `selectedTeamID` and sent to the API. Nothing
local validates it, normalizes case, or maps a slug to an ID.

**The resolver already exists in the same binary.** `cmd/ox/team_discovery.go:175`:

```go
func resolveTeamByQuery(projectRoot, query string) *enrichedTeam {
	q := strings.ToLower(strings.TrimSpace(query))
	// pass 1: exact slug match
	// pass 2: exact team ID match
	// pass 3: case-insensitive name match
}
```

Three passes, covering exactly the three vocabularies a user might reasonably type. It backs
`ox team show`, `ox team open`, and friends. `ox init --team` is the one team-taking surface
that does not use it — and it is the surface where being wrong is most expensive, because it
is the one that writes config and installs files.

## Why the failure is worse than a bad error message

The rejection arrives from the server, after `ox init` has already modified the working tree.
The failure therefore runs rollback, which currently deletes agent hook files it did not
create — #854. A typo in a CLI argument can cost a config file. That deletion is tracked
there; this report is only about the argument's vocabulary.

There is also no offline path: because validation is server-side, `--team` cannot be checked
before the run begins, even though the CLI can already enumerate the user's teams.

## Suggested direction

Happy to follow whatever shape you prefer.

- **Resolve `--team` through `resolveTeamByQuery` before registering.** Accepts slug, ID, or
  name, matching every other `ox team` surface, and lets a wrong value fail *before* the tree
  is touched.
- **On no match, fail locally with the candidates listed.** `Unknown team "platform". Your
  teams: Platform (platform, team_abc123), Research (research, team_xyz789).` The data is
  already in hand.
- **Update the help text** to whatever the flag ends up accepting — `team name, slug, or ID`
  if the resolver is adopted.
- Worth deciding explicitly whether a name that matches no team should ever reach the server.
  Today it does, which is why the error is phrased in the server's vocabulary rather than the
  CLI's.

If the current behaviour is deliberate — `--team` intentionally being a thin server
passthrough — then the fix is smaller: correct the help text, and say in it that the value
must match exactly and is case-sensitive.

## Environment

`ox 0.14.3` (brew, `sageox/tap`) and a local build of `main` @ `19fac267`, linux/amd64,
go1.26.5. Three-team account.

deepkawal:
Closes #855.

## What broke

`--team` is documented as taking a team ID. It also accepts a team **name**, case-sensitively. It rejects the **slug** that `ox team list` prints and that every other `ox team` subcommand accepts.

```
$ ox init --team platform
Error: failed to register with SageOx API: HTTP 400 from
https://sageox.ai/api/v1/repo/init: {"success":false,"error":"team name not found: \"platform\""}

$ ox init --team Platform
✓ Registered with SageOx (team: team_abc123)
```

Same team. Which of the three vocabularies you supply was not knowable from the help text, which said `team ID to associate this repo with`.

- **Where:** `cmd/ox/init.go:412-415` — the flag's raw string went straight to `selectedTeamID` and on to the API. Nothing local validated it, normalized case, or mapped a slug to an ID.
- **Why it costs more than a bad error message:** the rejection arrives from the server, *after* init has already modified the working tree. The failure therefore runs rollback — which currently deletes agent hook files it did not create (#854). A typo in a CLI argument can cost a config file.

**`ox init` was the outlier, not the rule.** `ox invite --team` already resolves the flag through a resolver and falls back to the raw value when it finds nothing (`cmd/ox/invite.go:472-479`), and the daemon's own IPC type documents slug as the *"kebab-case selector users may pass to `--team`"* (`internal/daemon/ipc.go:305`). Init was the one team-taking surface that skipped resolution — and the one where being wrong is most expensive, because it is the one that writes config and installs files.

## What this PR ships

- **Adds** `resolveTeamMembership(teams, query)` in `cmd/ox/team_discovery.go`, directly beneath `resolveTeamByQuery`, using the identical pass order: exact slug → exact team ID → case-insensitive name.
- **Adds** `fetchTeamMemberships()` and `formatTeamCandidates()` alongside it.
- **Changes** the `--team` branch to resolve before registering, and to fail locally — listing the user's teams — when nothing matches.
- **Changes** the flag help to `team to associate this repo with (name, slug, or ID)`.
- **Adds** `unknownTeamError()`, which distinguishes an account with teams from one with none.
- **Adds** 5 tests, 10 table cases.

A failed lookup now reads:

```
Error: unknown team "platfrm"

Your teams: Platform (platform, team_abc123), Developer Experience (dx, team_xyz789)
```

### Three decisions worth flagging

**Resolves against the API list, not `resolveTeamByQuery`.** The issue suggested reusing `resolveTeamByQuery`, but that answers from *locally cloned* team contexts. That is correct for `ox team show`, and wrong here: at init time the repo may have no local team data at all, which is exactly the fresh-machine case where `--team` is most likely to be used. This resolves against `TeamMembershipsFromRepos()` — the authoritative list init already fetches on the picker path. The two resolvers are adjacent in the file and share a pass order so the same string resolves to the same team on either path.

**A failed fetch falls through with the raw value** rather than erroring, so a degraded network cannot make `--team` unusable. Only a *successful* fetch that matches nothing is treated as a typo.

**An unmatched value now fails locally instead of reaching the server** — this is the one deliberate behavior change, and it is the part most worth your opinion. A team absent from `TeamMembershipsFromRepos()` would now be rejected locally where it previously reached the API. If you would rather `--team` stay a thin passthrough, the same change works with a warning in place of the error and I am happy to switch it — the resolution and the help text are useful either way.

**Where this lands relative to file creation.** The first unconditional tree write is `os.MkdirAll(sageoxDir)` at `init.go:486`, under `=== REPOSITORY SETUP ===`, so on a repo with commits this validation runs before init creates anything. One case is not covered: in an *empty* repo, `ensureInitialCommit` (`init.go:320`) has already written and committed `.sageox/README.md` by this point. I reproduced it against this branch — the run fails locally and writes nothing further, but the seed commit still lands:

```
$ ox init --team no-such-team-xyz --agents codex
✓ Created initial commit for empty repository
Using endpoint: sageox.ai
Error: unknown team "no-such-team-xyz"
...
$ git log --oneline
7d2a60d Initialize SageOx configuration
```

Closing that gap means hoisting endpoint selection and the auth gate above `ensureInitialCommit`, which reorders prompts and network calls in `runInit` — too much to ride along here, so it is filed separately as **#857** (with the reproduction, plus two smaller things the same repro turned up). Raised by review; the code comment points at the issue.

## Maintainer review round

`rsnodgrass` reviewed after CodeRabbit's three findings above were addressed and the PR sat clean, and found six more — three blocking, three non-blocking — all in `cmd/ox/team_discovery.go`, and all in code this diff had already added but nothing here had walked back out of. Addressed in `0d8064eb`:

| # | Finding | Blocking | Fix |
|---|---|---|---|
| 1 | `formatTeamCandidates` rendered `t.Name`/`t.Slug`/`t.ID` raw — all three are server-supplied, and this message fires on a typo, the path a user is least expecting output from | Yes | Routed through `cli.SanitizeTerminalText`, the same guard `renderTeamShow` and the invite path already apply to these fields. List capped at `maxTeamCandidates` (10) so a mistyped `--team` in CI can't dump the whole org chart. |
| 2 | First-match-wins over `TeamMembershipsFromRepos()`'s derived list, whose order is undefined — Go randomizes map iteration per process | Yes | `TeamMembershipsFromRepos()` now sorts the derived path (by ID, then Name). Independently, `resolveTeamMembership` no longer takes the first match — it collects every match from the first pass that hits anything, and `resolveTeamFlag` turns more than one into an `ambiguousTeamError` naming the candidates, instead of picking one. |
| 3 | `RepoInfo.TeamID` is `omitempty`, so a derived membership can have `ID == ""`; matching one silently drops it from `req.Teams` while `cfg.TeamName` and the success line still name it — the durable lie | Yes | New `usableTeams()` drops empty-ID memberships before any resolution pass runs. |
| 4 | `if reposResp == nil` in `fetchTeamMemberships` is unreachable — `GetRepos()` never returns `(nil, nil)` — and two stacked, contradictory doc comments were built on it | No | Guard removed; replaced with one accurate doc comment. |
| 5 | An unusable token degraded silently here, where the picker path ten lines away warns visibly for the identical condition | No | Now logs `slog.Warn` and calls `cli.PrintWarning` before falling through to the unauthenticated request, matching the sibling path. |
| 6 | Slug and name passes were case-insensitive, the ID pass was exact — a case-variant of a real ID fell through to the name pass and could match the wrong team | No | All three passes in `resolveTeamMembership` now use `strings.EqualFold`. |

Each fix has a corresponding test in `cmd/ox/team_discovery_test.go` / `internal/api/repos_coverage_test.go` — table-driven, including an ANSI/OSC-injection case for #1 and an at-cap boundary case for the candidate list. Full detail is in the six inline replies below.

**One more thing this round surfaced, left alone:** `resolveTeamByQuery` — the sibling resolver a few functions up in this same file, used by `ox team show` — has the identical case asymmetry on its ID pass (`t.TeamID == query`, exact and case-sensitive, while its own slug and name passes are case-insensitive). It's untouched by this PR; flagging rather than fixing it here since it's a different code path with its own review history — happy to file it separately or fold it in here, whichever you'd prefer.

### Verification after rebase (2026-09-17)

Rebased onto `origin/main` (`8c8993eb`). `git range-diff` against the previously pushed tip confirms the only semantic delta from the rebase itself is the `teamNameForID` removal already noted above — nothing else drifted.

| Check | Result |
|---|---|
| `gofmt -l` (changed files) | clean |
| `go vet ./cmd/ox/... ./internal/api/...` | clean |
| `make lint` | 0 issues |
| `make test-preflight` (lint + full + slow, `-race`) | **32,061 tests, 0 failed** (22,196 fast/full + 9,865 slow-tag) |
| `internal/ledger` (known map-iteration-order flake, see #935/#936) | passed clean this run |
| Repo-mutation check (`git status`, `AGENTS.md` diff) | clean |
| Mutation test on the sanitization guard | removed it → `TestFormatTeamCandidates_SanitizesServerText` goes red; restored → passes |

## Test Plan

| Check | Command | Result |
|---|---|---|
| New tests | `go test ./cmd/ox/ -run 'TestResolveTeamMembership\|TestFormatTeamCandidates\|TestUnknownTeamError'` | all pass |
| **Red-first proof** | removed the slug pass | **2 fail** — `slug that the name does not contain`, `slug wins over a name that collides with it` |
| **Red-first proof** | removed the empty-list branch | `TestUnknownTeamError` fails |
| Restore | reverted the stub | all pass |
| Package suite | `go test ./cmd/ox/` | ok |
| Full suite | `make test` | **19,131 tests, 999 skipped, 0 failed** |
| Lint | `make lint` | **0 issues** |
| Formatting | `gofmt -l` on changed files | clean |

The table's collision case (`dx` as one team's slug and another's name) pins the resolution order, and the non-derivable-slug case (`rnd` for `Research & Development`) is the one that fails without a real slug pass — for teams whose slug is just the lowercased name, the name pass alone would have hidden the bug.

**One pre-existing condition, flagged rather than hidden:** `cmd/ox/code.go` and `cmd/ox/code_test.go` are `gofmt`-dirty on clean `origin/main`. Both are untouched by this PR — I verified the dirt reproduces on `origin/main` with this change absent.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **Improvements**
  * The `--team` option now accepts team names, slugs, or IDs.
  * Team values are trimmed and matched without regard to letter case.
  * Matching prioritizes slugs, then IDs, then names.
  * Ambiguous or unknown team values now produce clear errors and available options when membership data is available.
  * Team selection remains available for downstream handling when memberships are unavailable or cannot be retrieved.
  * Team options are presented in a consistent order.


deepkawal:
Thanks — all three verified against the code. Two are fixed here; the third is correct but I have deliberately left it out of this PR.

**1. Long help text (`init.go:129`) — fixed.** Valid catch: I updated the flag's short help and missed the long description, which left the two disagreeing. It now reads `Use --team to specify a team by name, slug, or ID`.

**2. Unknown team after a successful empty response — fixed.** Correct, and the distinction matters more than the wording suggested: the previous shape treated *"could not reach the API"* and *"reached it, and you belong to no teams"* identically. Only the first is inconclusive. A successful fetch is now authoritative regardless of length, and an account with no teams gets the route that works instead of an empty candidate list:

```
unknown team "no-such-team": this account belongs to no teams

Run ox init without --team to create one
```

`TestUnknownTeamError` covers both branches and asserts the empty case does not render a blank `Your teams:` line.

**3. Validate before `ensureInitialCommit` — correct, reproduced, and filed as #857.**

The ordering is as you describe: `ensureInitialCommit` is at `init.go:320`, team resolution at `408`, so in a repo with no commits `.sageox/README.md` is written and committed before validation runs. I reproduced it rather than take it on reading, and the repro turned up two things worth more than the ordering itself:

- Rollback removes sixteen files and **misses `.sageox/README.md`** — it survives on disk and in the tree, because it is written before the tracker exists. `rollback()` also runs no git command at all, so the commit is untouchable by it. The run ends with `git status` **clean**, so nothing in the repo records that anything failed.
- The `-c` identity flags **override rather than fall back**, contrary to the doc comment at `init.go:167-168`. On the empty-repo path the root commit is authored `SageOx <ox@sageox.ai>` even when the developer has a git identity configured. I confirmed that in two fresh repos with different identities.

Full transcripts are in #857.

Two things bound it. It is a no-op once the repo has commits (`init.go:170`), so only empty repos are affected. And the first unconditional tree write is `os.MkdirAll(sageoxDir)` at `init.go:486`, under `=== REPOSITORY SETUP ===` — after this validation. So in the ordinary case the change does land before file creation, and the empty-repo path is the remaining gap.

I have left it because the fix is what you labelled it. Hoisting endpoint selection, the auth gate, and team resolution above `ensureInitialCommit` reorders user-facing prompts and network calls in `runInit`, and that deserves its own PR and its own review rather than riding along with a flag-resolution change.

Worth noting the gap predates this PR rather than being introduced by it: `--team` was previously validated by the server, i.e. after *every* file had been written and staged. This moves validation earlier, just not all the way to the first byte.

I have corrected the PR description, which overclaimed on exactly this point, and the code comment now points at #857. Glad to take the reordering as a follow-up PR once you have a view on which of the three directions in that issue you prefer — or to fold it in here instead, if you would rather have it in one go.

deepkawal:
Fixed in c5baa9b, along with a correction to my own previous round.

**The regression was mine.** In 6bd931b I collapsed "reached the API, list is empty" into "authoritative, therefore reject". That was wrong, and your reading of it is exactly right: `fetchTeamMemberships` returns `(nil, nil)` when `GetRepos` yields no response body, so that shape made a missing body hard-fail a valid `--team` with *"this account belongs to no teams"*. The code comment I wrote asserting the empty list was authoritative was simply false. Thanks for catching it.

The three outcomes are now explicit, in one function with the rule stated in its doc comment:

| Fetch outcome | `--team` |
|---|---|
| `fetchErr != nil` | passed through — nothing is knowable |
| `len(teams) == 0` | passed through — no usable answer |
| `len(teams) > 0` | authoritative — an absent value is a typo |

`TestResolveTeamFlag_States` pins all five cases, including both empty shapes separately.

**One deliberate divergence.** You asked that `unknownTeamError` still fire for an authoritative empty list. I don't reject on empty at all, for two independent reasons.

The first is that "authoritative empty" isn't currently distinguishable at this layer. `(nil, nil)` for a bodyless response and a genuine zero arrive as the same value, so any rejection here would have to split them apart first. I've written that constraint into the doc comment so the next person to reach for a rejection sees why it isn't safe yet.

The second is that an unambiguous zero still shouldn't reject, because `TeamMembershipsFromRepos` *derives* memberships from repos of type `team-context` whenever the server doesn't populate `teams` — so an account with a real, freshly created team that has no team-context repo yet produces an empty list while genuinely having teams. And `init` already treats zero teams as continuable everywhere else: on the picker path `promptNoTeams` offers *"Continue (a new team will be created)"* and proceeds. Rejecting here would make `--team` the only surface on which zero teams is fatal.

The trade is asymmetric. Passing through costs a less precise server-side error for an account that truly has no teams. Rejecting costs a blocked `init` for one whose team context is merely still provisioning.

**On the explicit availability state**, I built it first — a `{Teams, Available}` struct — and reverted it. Red-first showed it was decoration: deleting the `!Available` guard passed the entire suite, because whenever `Available` is false the list is also empty and the next guard catches it. It added a field without adding a behavior any test could pin. The slice signature plus a documented rule is what the tests can actually hold onto.

---

**One correction to the thread.** The comment on validating `--team` before the first working-tree mutation is marked *"✅ Addressed in commit 6bd931b"*. It isn't. That finding is real, I reproduced it, and it is filed as #857 and deliberately left out of this PR — the fix reorders user-facing prompts and network calls in `runInit` and deserves its own review. Flagging it so the marking doesn't read as done to anyone scanning the thread.

cmd/ox/team_discovery_test.go
```diff
@@ -129,3 +132,133 @@ func TestResolveTeamByQuery_WhitespaceHandling(t *testing.T) {
...
+// TestResolveTeamFlag_States pins which fetch outcomes may reject --team: only a
+// NON-EMPTY membership list may. Every other outcome passes the trimmed value
+// through to the server, which is the authority on team names.
+func TestResolveTeamFlag_States(t *testing.T) {
+	teams := []api.TeamMembership{
+		{ID: "team_abc123", Name: "Platform", Slug: "platform"},
+		{ID: "team_xyz789", Name: "Research & Development", Slug: "rnd"},
+	}
+
+	t.Run("transport error passes the trimmed value through", func(t *testing.T) {
```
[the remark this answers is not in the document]
deepkawal: Taken — `49de927d`. Valid per AGENTS.md, and the repetition was real.

One thing I kept rather than folded away. Each case existed for a specific reason, and those reasons were the point of the test: the nil and empty-non-nil rows are deliberately distinct states, because `fetchTeamMemberships` returns `(nil, nil)` for a response with no body, and reading that as "this account has no teams" is the regression this file exists to prevent. So the table carries a `why` field, and it is used as the assertion message rather than sitting in a comment:

```go
require.NoError(t, err, tt.why)
```

A future regression then reports the rule it broke instead of just a value mismatch.

Case names and expectations are unchanged, and I verified the table still bites: deleting the `len(teams) == 0` guard from `resolveTeamFlag` fails exactly the two pass-through rows and leaves the other three green.

cmd/ox/team_discovery.go
```diff
@@ -204,6 +206,130 @@ func resolveTeamByQuery(projectRoot, query string) *enrichedTeam {
...
+	return reposResp.TeamMembershipsFromRepos(), nil
+}
+
+// formatTeamCandidates renders the user's teams for an error message, so a failed
+// --team lookup can show what would have worked rather than only what did not.
+func formatTeamCandidates(teams []api.TeamMembership) string {
+	parts := make([]string, 0, len(teams))
+	for _, t := range teams {
+		if t.Slug != "" {
+			parts = append(parts, fmt.Sprintf("%s (%s, %s)", t.Name, t.Slug, t.ID))
```
rsnodgrass, line 279: **Blocking — unsanitized server text to the terminal.**

`t.Name`, `t.Slug` and `t.ID` come straight from the API and go out raw. We have a canonical guard for exactly this, and `internal/cli/sanitize.go`'s doc comment names the class explicitly:

> Any string that originates outside this binary — a knowledge bubble's name/description/steering, **a team name**, an API error body — can carry ANSI CSI/OSC escape sequences... (with OSC 8 and OSC 52) smuggle hyperlinks and clipboard writes past the user.

`renderTeamShow` (`cmd/ox/team.go:504-525`) sanitizes precisely these three fields, and `invite.go:775` does the same. This is currently the only team-rendering path that skips it — and it fires on a typo, printing every team verbatim.

```go
if t.Slug != "" {
    parts = append(parts, fmt.Sprintf("%s (%s, %s)",
        cli.SanitizeTerminalText(t.Name),
        cli.SanitizeTerminalText(t.Slug),
        cli.SanitizeTerminalText(t.ID)))
    continue
}
parts = append(parts, fmt.Sprintf("%s (%s)",
    cli.SanitizeTerminalText(t.Name),
    cli.SanitizeTerminalText(t.ID)))
```

**Recommendation while you're here:** cap the list at ~10 with `and N more — run 'ox team list'`, so a typo in CI doesn't print the whole org chart.

cmd/ox/team_discovery.go
```diff
@@ -204,6 +206,130 @@ func resolveTeamByQuery(projectRoot, query string) *enrichedTeam {
...
 
+// resolveTeamMembership finds a team in an API-supplied membership list by slug,
+// team ID, or name, using the same resolution order as resolveTeamByQuery above.
+//
+// The two resolvers sit next to each other on purpose. resolveTeamByQuery answers
+// from locally cloned team contexts, which is right for `ox team show` but wrong for
+// `ox init`: at init time the repo may have no local team data at all. This one
+// answers from the authoritative membership list the API returns, and keeping the
+// pass order identical means the same string resolves to the same team on either path.
+func resolveTeamMembership(teams []api.TeamMembership, query string) *api.TeamMembership {
```
rsnodgrass, line 217: **Blocking — first-match-wins over a nondeterministically ordered list.**

Each pass returns on the first hit, with no check for a second. That is only well-defined if the input order is stable, and it isn't: `TeamMembershipsFromRepos()` (`internal/api/repos.go:97`) falls back to

```go
for _, repo := range r.Repos   // Repos is map[string]RepoInfo
```

and Go randomizes map iteration per process.

**Concrete failure, no attacker needed:** a consultant belongs to two orgs that each named a team `Platform`. `ox init --team platform` resolves to org A on Monday and org B on Tuesday — same command, same machine — and the repo's Ledger lands in the wrong client's tenant. Nothing in the output distinguishes the two runs.

This is the part that makes moving resolution client-side risky in kind. Previously the server owned the uniqueness constraints; now the client decides, and it doesn't know them.

**Recommendation** — either is fine, `(a)` is cheaper and fixes it for every consumer:

- **(a)** sort the fallback path in `TeamMembershipsFromRepos()` by ID before returning; or
- **(b)** collect all matches per pass and, on `len(matches) > 1`, return an error listing the candidates and asking for the unambiguous ID.

Ambiguity shouldn't resolve silently either way.

cmd/ox/team_discovery.go
```diff
@@ -204,6 +206,130 @@ func resolveTeamByQuery(projectRoot, query string) *enrichedTeam {
...
+	if fetchErr != nil {
+		slog.Debug("could not fetch teams to resolve --team; passing the value through", "error", fetchErr)
+		return trimmed, "", nil
+	}
+	if len(teams) == 0 {
+		slog.Debug("no team memberships reported; passing --team through unresolved", "team", trimmed)
+		return trimmed, "", nil
+	}
+	if match := resolveTeamMembership(teams, trimmed); match != nil {
+		return match.ID, match.Name, nil
```
rsnodgrass, line 328: **Blocking — `match.ID` can be empty, which silently drops the association while the output still claims it.**

`RepoInfo.TeamID` is `json:"team_id,omitempty"` and `StableID()` returns it, so `TeamMembership.ID` can be `""` on the fallback path in `TeamMembershipsFromRepos()`. This returns it unchecked, and then:

- `init.go:833` — `if selectedTeamID != ""` is **false**, so `req.Teams` is never set and the server takes its "no team specified" path.
- `init.go:914` — `cfg.TeamName` is written anyway.
- The success line prints `formatTeamLabel(selectedTeamName, resp.TeamID)`.

Net: `.sageox/` config **and** the success message both name a team the repo was never registered to. That's a durable lie that `ox status` and `ox doctor` will later read back.

This was impossible before the PR, since `--team X` always produced a non-empty `selectedTeamID`.

**Recommendation:** treat an empty ID as unresolved — skip such candidates inside `resolveTeamMembership` rather than special-casing here, so every caller benefits.

cmd/ox/team_discovery.go
```diff
@@ -204,6 +206,130 @@ func resolveTeamByQuery(projectRoot, query string) *enrichedTeam {
...
+func fetchTeamMemberships() ([]api.TeamMembership, error) {
+	teamClient := api.NewRepoClient()
+	if token, err := auth.EnsureValidToken(300); err == nil && token != nil && token.AccessToken != "" {
+		teamClient.WithAuthToken(token.AccessToken)
+	}
+	reposResp, err := teamClient.GetRepos()
+	if err != nil {
+		return nil, err
+	}
+	if reposResp == nil {
```
rsnodgrass, line 267: **Non-blocking — this branch is unreachable, and two doc comments are built on it.**

`GetRepos()` (`internal/api/repos.go:179-238`) has exactly two exits: `(nil, err)` on any failure, or `(&reposResp, nil)` on success. `reposResp` is a local struct value, so `&reposResp` is never nil — even for an empty or `null` body, since `json.Unmarshal` into a struct pointer treats `null` as a no-op rather than an error. **`GetRepos()` cannot return `(nil, nil)`.**

So this guard is dead, and the `(nil, nil)` reasoning in both this function's comment and `resolveTeamFlag`'s describes a contract the code doesn't have.

Your *conclusion* is still right — an empty list must not be read as an authoritative zero, and passing through matches how `promptNoTeams` treats zero teams elsewhere in `runInit`. It just needs a true reason.

**Recommendation:** drop the branch, and reduce the justification to something defensible, e.g. *"an empty result is not proof of zero teams — it also covers degraded or denied responses — so callers must not treat it as an authoritative zero."*

Also: this function currently carries **two stacked doc-comment blocks**, both opening `fetchTeamMemberships returns...`, and they disagree — the first says an unauthenticated request is attempted, which the `ErrUnauthorized` path contradicts. Worth collapsing to one.

cmd/ox/team_discovery.go
```diff
@@ -204,6 +206,130 @@ func resolveTeamByQuery(projectRoot, query string) *enrichedTeam {
...
+// is attempted unauthenticated and the caller decides what an empty result means.
+// fetchTeamMemberships returns the teams the API reports for the current user.
+//
+// Note that it returns (nil, nil) when GetRepos yields no response body: reachable,
+// but no answer. That is indistinguishable here from an account that genuinely has
+// no team contexts, so callers must not read an empty result as a count of zero.
+// resolveTeamFlag is where that rule is enforced.
+func fetchTeamMemberships() ([]api.TeamMembership, error) {
+	teamClient := api.NewRepoClient()
+	if token, err := auth.EnsureValidToken(300); err == nil && token != nil && token.AccessToken != "" {
```
rsnodgrass, line 260: **Non-blocking — the token error is swallowed, and the sibling branch ten lines below shows what we normally do.**

The picker path at `init.go:419-432` handles this same fetch carefully:

```go
token, tokenErr := auth.EnsureValidToken(300)
if tokenErr != nil {
    slog.Debug("token ensure failed after auth gate passed", ...)
}
if token != nil && token.AccessToken != "" {
    teamClient.WithAuthToken(token.AccessToken)
} else {
    slog.Warn("no usable token for team fetch despite passing auth gate", ...)
}
```

and on a fetch error it emits a `cli.PrintWarning` plus an interactive "Continue without team selection?" confirm.

Folding all of that into `err == nil && token != nil && ...` means the `--team` path degrades **silently** where the picker path degrades **visibly** — after the auth gate at `init.go:397` has already proved the user is authenticated. The outcome is benign (no token → server 401s → pass-through), but a user with a broken token gets no signal at all.

**Recommendation:** match the sibling — warn on an unusable token, and surface the fail-open with `cli.PrintWarning` rather than `slog.Debug` only, so a silently-unvalidated `--team` is visible.

cmd/ox/team_discovery.go
```diff
@@ -204,6 +206,130 @@ func resolveTeamByQuery(projectRoot, query string) *enrichedTeam {
...
+	// swallow queries that should have fallen through to ID or name.
+	for i, t := range teams {
+		if t.Slug != "" && strings.ToLower(t.Slug) == q {
+			return &teams[i]
+		}
+	}
+
+	// pass 2: exact team ID match
+	for i, t := range teams {
+		if t.ID == trimmed {
```
rsnodgrass, line 234: **Non-blocking — case handling is inconsistent across the three passes.**

Slug uses `strings.ToLower`, name uses `strings.EqualFold`, and ID here is exact and case-sensitive. Exact-match on ID is defensible on its own — IDs are opaque generated tokens, not something people free-type — but the asymmetry means any case-variant of a real team ID falls *through* this pass and can be picked up by the case-insensitive **name** pass below.

Slugs are server-derived lowercase ASCII, so `ToLower` vs `EqualFold` has no practical Unicode exposure there; team **names** are the field that can carry non-ASCII, and those already use the more correct `EqualFold`.

**Recommendation:** unify on `EqualFold` throughout for defensiveness. Cheap, and it removes the fall-through entirely.

rsnodgrass (changes requested):
Thanks for this — the write-up is genuinely excellent. You found a real inconsistency, you checked how the sibling surfaces behave before proposing a fix, you reproduced the empty-repo edge case and filed it separately as #857 instead of scope-creeping, and you flagged the one deliberate behavior change and asked for a ruling on it. That is exactly the shape we want.

Requesting changes on four things. Three are small and local. One is a test-harness gap that matters more than it looks.

## Blocking

### 1. The new code path has no end-to-end coverage — and the existing E2E tests silently exercise the *old* behavior

This is the big one.

- `cmd/ox/e2e_harness_test.go`'s stub server switches on `r.URL.Path` and handles only `/api/v1/repo/init`. `reposPath = "/api/v1/cli/repos"` is **not** stubbed.
- Four existing tests in `init_e2e_test.go` call `withInitFlags(t, env.TeamID)` and drive `runInit()` for real.
- After this PR, each of those hits `fetchTeamMemberships()` → unhandled route → error → `fetchErr != nil` → the **pass-through branch**. They all keep passing, and every one of them is proving pre-PR behavior.

So the board goes green while nothing verifies the change. Your unit tests over `resolveTeamMembership` / `resolveTeamFlag` are good and worth keeping, but the wiring — a real token, a real `GetRepos()`, the resolved ID actually reaching `req.Teams` and the name reaching `cfg.TeamName` — is untested.

The harness already has everything needed. What we'd want:

- Stub `/api/v1/cli/repos` in `newOxE2E` to return `env.TeamID` alongside a slug.
- One test: `--team <slug>` end-to-end, asserting `/api/v1/repo/init` received the **resolved** team ID.
- One test: `--team <unknown>` asserting `.sageox/` is **never created** — that's your headline claim, and it currently has no proof.

### 2. `formatTeamCandidates` writes unsanitized server text to the terminal

We have a canonical guard for exactly this data class, and its doc comment names team names specifically:

> Any string that originates outside this binary — a knowledge bubble's name/description/steering, **a team name**, an API error body — can carry ANSI CSI/OSC escape sequences... (with OSC 8 and OSC 52) smuggle hyperlinks and clipboard writes past the user.

`renderTeamShow` (`cmd/ox/team.go:504-525`) sanitizes precisely `name`, `teamID` and `slug`. `invite.go:775` does the same. `formatTeamCandidates` is currently the only team-rendering path that skips it — and it fires on a typo, printing every team verbatim.

Wrap each field in `cli.SanitizeTerminalText`.

### 3. Resolution is nondeterministic when two teams collide

`TeamMembershipsFromRepos()` (`internal/api/repos.go:97`) falls back to `for _, repo := range r.Repos`, and `Repos` is a `map[string]RepoInfo`. Go randomizes map iteration per process. Combined with return-on-first-hit and no duplicate check, an account in two orgs that each named a team "Platform" gets a different answer on different runs of the same command.

This is the part that makes moving resolution client-side risky in kind: the server owned uniqueness before, and the client doesn't know those rules. Please either sort the fallback path by ID, or treat a pass yielding more than one hit as an error listing the candidates. Ambiguity shouldn't resolve silently.

### 4. A resolved-but-empty team ID drops the association while the output still claims it

`RepoInfo.TeamID` is `json:"team_id,omitempty"` and `StableID()` returns it, so `TeamMembership.ID` can be `""` on the fallback path. `resolveTeamFlag` returns `match.ID` unchecked. Then:

- `init.go:833` — `if selectedTeamID != ""` is false, so `req.Teams` is never sent and the server takes the no-team path.
- `init.go:914` — `cfg.TeamName` is still written.
- The success line prints `formatTeamLabel(selectedTeamName, resp.TeamID)`.

Net: the config file and the success message both name a team the repo was never registered to. That was impossible before this PR, since `--team X` always produced a non-empty ID. Treat `match.ID == ""` as unresolved.

## Non-blocking

- **`if reposResp == nil` is dead.** `GetRepos()` returns `&reposResp, nil` on success and `nil, err` otherwise — `(nil, nil)` can't happen. The conclusion built on it (pass through on empty) is still right, but the stated reason describes a contract the code doesn't have. Also, `fetchTeamMemberships` has two stacked doc-comment blocks that contradict each other on whether an unauthenticated request is attempted.
- **The `EnsureValidToken` error is swallowed.** The picker branch ten lines below (`init.go:419-432`) captures it, `slog.Warn`s when the token is unusable, and on fetch failure emits a `cli.PrintWarning` plus an interactive confirm. The new path does none of that, so `--team` degrades silently where the sibling degrades visibly. Worth matching.
- **Case-handling asymmetry.** Slug uses `ToLower`, name uses `EqualFold`, ID is exact. Unify on `EqualFold` for defensiveness.
- **Unbounded candidate list.** Cap at ~10 with "and N more — run `ox team list`", so a typo in CI doesn't print the whole org chart.

## On the ruling you asked for

You asked whether `--team` should fail locally or stay a passthrough. Two things we turned up while reviewing:

- **`ox invite --team` already made the opposite call, deliberately** (`cmd/ox/invite.go:471-479`): *"Unknown locally is not an error: this machine may simply never have synced that team. Let the server rule on it."* So the PR matches every other team-taking surface on **vocabulary**, but inverts the **failure policy**. Worth knowing, since consistency was part of the argument.
- A token authorized to register into a team, but whose `/api/v1/cli/repos` list doesn't include that team, would now be rejected client-side for a value the server would have accepted.

Our lean is **warn and pass through** on a non-match, matching `invite`. You keep all the resolution value — the slug works, the name works, the picker vocabulary works — and the server stays the authority on what's valid. It also makes #3 and #4 far less dangerous, since a bad client-side resolution stops being able to silently pick a wrong team.

That said, the typo-guard you were originally chasing is real value and we'd rather not lose it entirely. If you want to keep a local failure for the unambiguous case, we're open to it — say which way you'd like to go and we'll back it.

## CI

CI had never actually run on this PR — all three workflows were sitting at `action_required` because it's from a fork. That's on us, not you. I've approved them, and there is now a real board.

`lint` and `Security review` pass. **`test` fails**, on one thing:

```
Generated CLI docs are stale. Run 'make docs' and commit the result.
make: *** [Makefile:613: docs-check] Error 1
```

You changed the `--team` flag help and the long help in `init.go`, which `docs/reference/init.mdx` is generated from. The diff `make docs-check` reports is exactly your two intended wording changes, so this is purely a regeneration step:

```
go build -o ox-tmp ./cmd/ox && ./ox-tmp docs --output docs/reference && rm ox-tmp
```

or just `make docs`, then commit the result. Nothing wrong with the change itself.

rsnodgrass:
Status check on this one so it does not sit — the change is wanted, it is just stalled on four mechanical things and nobody has said so out loud in nine days.

- **`test` is red for a boring reason.** It is not a test failure: `docs-check` caught that the generated CLI reference no longer matches the cobra definitions, because you (correctly) reworded the `--team` help. Fix is `make docs` and commit `docs/reference/init.mdx`.
- **The branch now conflicts with `main`.** It needs a merge from `main` — please merge rather than rebase, so the existing review threads stay anchored.
- **Three blocking threads are still open** on `cmd/ox/team_discovery.go` (unsanitized server text reaching the terminal · first-match-wins over a nondeterministically ordered list · empty `match.ID` silently dropping the association while the output still claims success). Those are the ones that gate the merge.
- The three non-blocking notes on the same file are yours to take or leave.

`lint`, `coverage`, and `codecov/patch` are all green, so once those four are handled this should go straight through. Ping here when it is ready and it will get re-reviewed promptly.

deepkawal:
This is on my windows machine and already fixed , will push tonight after
doing a final check

cmd/ox/team_discovery.go
deepkawal: Fixed in `0d8064eb`. `formatTeamCandidates` now runs `t.Name`, `t.Slug`, and `t.ID` through `cli.SanitizeTerminalText` before they reach the terminal — the same guard `renderTeamShow` and the invite path already apply to these exact fields. Also capped the list at `maxTeamCandidates` (10), so a mistyped `--team` in CI points at the fix rather than printing the whole org chart.

Covered by `TestFormatTeamCandidates_SanitizesServerText` (asserts the ANSI escape and BEL bytes are stripped and the printable text survives verbatim) and `TestFormatTeamCandidates_CapsTheList`. Mutation-tested: removed the sanitize calls, confirmed the test goes red, restored them.

cmd/ox/team_discovery.go
deepkawal: Fixed in `0d8064eb`, two ways.

`TeamMembershipsFromRepos()` (`internal/api/repos.go`) now sorts its derived path by ID then Name, so the map iteration randomization you flagged no longer reaches any caller. Covered by `TestTeamMembershipsFromRepos_DerivedOrderIsStable`, which calls it 50 times and asserts the order holds every time — a single call could agree with sorted order by chance, so the test doesn't stop at one.

Independent of the sort: `resolveTeamMembership` no longer takes the first match at all — it collects every match from the first pass that hits anything, and `resolveTeamFlag` turns more than one into an `ambiguousTeamError` naming the candidates. So a name or slug two teams share is rejected, not silently picked, even before considering ordering. Covered by `TestResolveTeamMembership_Ambiguous` and the ambiguous case in `TestResolveTeamFlag_States`.

cmd/ox/team_discovery.go
deepkawal: Fixed in `0d8064eb`. New `usableTeams()` drops any membership with an empty ID before any resolution pass runs, so the ID-less shape you traced (reachable via `RepoInfo.TeamID`'s `omitempty`) can never be matched by slug or name — closing exactly the durable-lie failure mode you described (empty ID silently drops `req.Teams` while `cfg.TeamName` and the success line still name the team).

Covered by `TestResolveTeamMembership_EmptyTeamList` and the two "ghost" cases in `TestResolveTeamFlag_States` ("a list of only ID-less teams passes through instead of rejecting" and "an ID-less team never resolves, even matching by name").

cmd/ox/team_discovery.go
deepkawal: Fixed in `0d8064eb`. Replaced the two stacked, contradictory doc comments on `fetchTeamMemberships` with one, and removed the `if reposResp == nil` guard along with the dead-branch reasoning it rested on — confirmed `GetRepos()` cannot return `(nil, nil)` before deleting it, per your trace.

cmd/ox/team_discovery.go
deepkawal: Fixed in `0d8064eb`. `fetchTeamMemberships` now matches the picker path's degrade behavior: on an unusable token it logs `slog.Warn` and calls `cli.PrintWarning` (visibly, not silently) before falling through to the unauthenticated request — same as the sibling ten lines away in `init.go` you pointed at.

cmd/ox/team_discovery.go
deepkawal: Fixed in `0d8064eb`. All three passes in `resolveTeamMembership` now use `strings.EqualFold`, including the ID pass, closing the fall-through you described. Covered by the "team ID is case-insensitive like every other pass" case in `TestResolveTeamMembership_TableDriven`.

One more thing this made me check: `resolveTeamByQuery` — the sibling a few functions up in this same file, used by `ox team show` — has the identical asymmetry on its own ID pass (`t.TeamID == query`, exact and case-sensitive, while its slug and name passes are case-insensitive). It's untouched by this PR since it's a different code path with its own review history; noted in the PR description, and happy to file it separately or fold it in here, whichever you'd rather.

PR #954, opened 2026-09-15
fix(daemon): prevent settling after accumulator stop

### LINKED ISSUE #953: ChangeAccumulator.settle() ignores the stopped flag, so a callback can fire after Stop()
`ChangeAccumulator.Stop()` sets a `stopped` flag, but `settle()` never reads it. A settle timer that fires concurrently with `Stop()` therefore still invokes `onSettled` — the exact failure `TestChangeAccumulator_OnSettledNotCalledAfterStop` says it prevents ("stale callback fires during daemon shutdown").

## The gap

`internal/daemon/change_accumulator.go` — `stopped` is written once and read once, and the read is in the wrong place:

| Line | Use |
|---|---|
| 53 | `stopped bool` field |
| 73 | read by `AddChange` |
| 181 | written by `Stop()` |
| 139–155 (`settle`) | **never read** |

`resetSettleTimer` schedules `time.AfterFunc(a.settlePeriod, func() { a.settle() })`. `Timer.Stop()` returns false once the timer has fired and cannot retract a callback that is already running or queued, so `Stop()` racing a firing timer leaves `settle()` free to reach:

```go
if a.onSettled != nil {
    go a.onSettled()
}
```

`settle()` already takes `a.mu`, and `Stop()` holds the same mutex while setting `stopped`, so an early return on `a.stopped` closes the window without new synchronization.

## How it surfaced

The `coverage` job failed on an unrelated PR (#951, which touches only `internal/ledger`):

```
--- FAIL: TestChangeAccumulator_OnSettledNotCalledAfterStop (0.34s)
    change_accumulator_test.go:211:
        Error: Not equal:
               expected: 0
               actual  : 1
        Messages: callback should not fire after Stop()
```

## What I could and could not reproduce

- `go test -run TestChangeAccumulator_OnSettledNotCalledAfterStop -count=200` — passes.
- Two full `go test ./internal/daemon/ -cover` runs — both pass.
- So it does not reproduce on an idle machine; it needs the loaded CI runner (21,686 tests, coverage instrumentation, `t.Parallel()` siblings).

**Honest limit:** I cannot tell from the CI log which of two things failed, and the fix differs:

1. **The race above** — the timer fired concurrently with `Stop()` and `settle()` ran anyway. A real product bug.
2. **Test timing** — the test sleeps 20ms against a 100ms settle period before calling `Stop()`. On a loaded runner that 20ms can overrun 100ms, so the timer fires *legitimately* before `Stop()` and the product is behaving correctly.

Both are worth fixing, and they are independent:

- For (1): return early from `settle()` when `a.stopped`.
- For (2): stop encoding the ordering in a sleep. Drive the settle deterministically — e.g. inject a clock, or assert on an observable transition rather than racing a wall-clock window.

Fixing only (1) leaves the test able to flake via (2), because in that path `settle()` completes before `stopped` is ever set.

## Why this is a separate issue

`internal/daemon` is a different subsystem from the PR that surfaced it, and per `CLAUDE.md` scope discipline an unrelated fix should not ride along in a ledger PR.

keppy:
## What broke

`ChangeAccumulator.Stop()` sets `stopped` and stops the settle timer, but `time.Timer.Stop()` cannot retract a callback that has already been dispatched. When that callback later entered `settle()`, it ignored `stopped`, published the pending changes, and launched `onSettled` during daemon shutdown.

## What this PR ships

- Check `stopped` inside `settle()` while holding the accumulator mutex, before publishing pending changes or dispatching the callback.
- Replace the scheduler-dependent sleep in the shutdown regression test with `testing/synctest` virtual time.
- Cover both a timer canceled before it fires and a timer callback that enters `settle()` after `Stop()`.
- Clarify that `Stop()` prevents future settling but does not join callbacks dispatched by a settle that already completed.

Closes #953.

## Test Plan

- **Red-first:** without the `stopped` guard, the late-settle case publishes one change and invokes `onSettled` once.
- `go test -race -short -tags short internal/daemon/change_accumulator.go internal/daemon/change_accumulator_test.go -count=1`
- `go test -race -short -tags short internal/daemon/change_accumulator.go internal/daemon/change_accumulator_test.go -run '^TestChangeAccumulator_' -count=50`
- `go test internal/daemon/change_accumulator.go internal/daemon/change_accumulator_test.go -run '^TestChangeAccumulator_OnSettledNotCalledAfterStop$' -count=200`
- `go vet internal/daemon/change_accumulator.go internal/daemon/change_accumulator_test.go`
- `golangci-lint v2.11.4 run -c .config/golangci.yml --allow-parallel-runners ./internal/daemon` — 0 issues

The repository-wide Windows test run is currently blocked by an existing Unix-only `syscall.Flock` use in `internal/doctor/autofix/skills_inventory_drift_test.go`. GitHub Actions will run the full Linux suite.

---
<details>
<summary>Checklist (expand if needed)</summary>

- [x] Tests added
- [x] CHANGELOG entry N/A — internal shutdown race fix
- [x] Breaking change N/A
- [x] Security review N/A — this change does not touch a security-review path

</details>

---
<a href="https://app.blacksmith.sh/sageox/codesmith/ox/pr/954"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792099704&installation_model_id=433550&pr_number=954&repository=sageox%2Fox&return_to=https%3A%2F%2Fgithub.com%2Fsageox%2Fox%2Fpull%2F954&signature=e4d660f9ff79b849d2c95b02de904b95c9ff6483b0e08ba840ca86b1eb015a2d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented pending changes from being processed after the accumulator is stopped.
  * Ensured settlement callbacks do not run after stopping, including callbacks that were already dispatched.

* **Documentation**
  * Clarified that stopping does not wait for callbacks already in progress.


rsnodgrass:
Thanks @keppy for the contribution! Reviewing now.

rsnodgrass (approved):
Approving. Verified independently rather than taking the test at its word.

**The bug is real.** `resetSettleTimer` schedules `settle` via `time.AfterFunc`. Once that callback is dispatched it blocks on `a.mu`, so a concurrent `Stop()` can take the lock first, set `stopped`, and call `settleTimer.Stop()` — which is a no-op on an already-fired timer. The callback then acquires the lock and publishes into `settled` and dispatches `onSettled` after shutdown. The guard closes it at the only place that can decide correctly: inside the same critical section that sets the flag.

**Red-first confirmed locally.** Removing just the five-line guard fails exactly the step that names the claim, and nothing else:

```
--- FAIL: TestChangeAccumulator_OnSettledNotCalledAfterStop/late_settle_after_stop
    expected: 0
    actual  : 1
    Messages: callback should not fire after Stop()
    Should be empty, but was [{src/main.go modified false ...}]
```

Restored, `go test ./internal/daemon/ -race -count=1` is green (257s, full package).

**On the test shape** — the two things I checked hardest, both good:

- The `lateSettle` branch calling `acc.settle()` directly is the right call. A dispatched-but-not-yet-entered timer callback is not reachable through the public API, and modeling it with real sleeps is exactly the flake this PR is replacing. The comment says why it is there, so the next reader will not "clean it up."
- `DrainSettled()` being asserted empty alongside the callback count matters more than it looks. The callback is the visible symptom; the leaked entry in `settled` is the one that would have survived a narrower fix.

`testing/synctest` already has precedent here (`internal/ledger/murmur_sparse_test.go`), and `go.mod` is on 1.26, so no new floor.

The `Stop()` doc change is worth calling out separately: it is now honest about not joining in-flight callbacks, which is the actual contract. A comment that had quietly stopped being true is its own small bug.

No security surface — the diff adds a bool read under an existing mutex; no IPC, exec, dependency, or credential path is touched.

Thanks for the clean first contribution, and for wiring it to #953.

keppy:
ty for the quick review, Ryan. Happy to take more of these if there's a backlog you want cleared. The synctest migration pattern would apply to a few other timer-dependent tests if that's useful.

PR #972, opened 2026-09-17
fix(daemon): reclaim abandoned draft sessions for anti-entropy

### LINKED ISSUE #966: Abandoned draft sessions are invisible to anti-entropy forever — draft guard runs before the staleness check, and --force-session-uploads doesn't force
## Summary

**Two of my sessions — 1151 and 628 entries of real work — sat unuploaded for 21 hours with a healthy daemon running the whole time, and `ox doctor --force-session-uploads` reported there was nothing to do.** They were invisible to anti-entropy, invisible to the force path, and produced no error anywhere. The only signal was `ox agent list` still showing `rec` against sessions whose Claude Code processes had been dead since the previous evening.

**Root cause:** in `SessionFinalizeHandler.detectInDir`, the draft guard runs *before* the staleness check, so a session whose process died while `draft: true` is skipped unconditionally — forever. `isStaleRecording` has exactly the right logic to catch this (dead PID ⇒ stale immediately), and it is unreachable for any draft.

Observed on ox `0.15.0 (260912-023348)` / commit `05043a01`, macOS 26.2.

## Observed

```
$ ox agent list
  ID        Type     Tokens   Cmds   Uptime  Rec  Whisper
  OxPcx9    claude    ~2.4K      1    21.8h  rec  48m ago   idle
  OxLtF2    claude    ~2.4K      1    20.7h  rec  14h ago   idle
```

Daemon uptime at the time: **23h 9m**. Not a dead-daemon case — it had been running and syncing every 60s the entire period.

Both sessions on disk:

```
2026-09-16T18-54-<me>-OxPcx9   draft=True  published=NO  entries=991
2026-09-16T20-00-<me>-OxLtF2   draft=True  published=NO  entries=447
```

Running the documented remedy, twice, ~15 min apart:

```
$ ox doctor --force-session-uploads
...
INFO session finalize detect complete ledger=/Users/<me>/.local/share/sageox/.../repo_<id> items=0
INFO force detect complete items_queued=0
INFO ForceDetect complete op=doctor queued=0
```

`items=0`. No warning, no error, no mention of the two sessions it just walked past. Meanwhile `ox agent <id> session recover` worked instantly on both:

```json
{ "success": true, "session_name": "2026-09-16T18-54-<me>-OxPcx9",
  "entry_count": 1151, "uploaded": true, "message": "session recovered via normal processing" }
{ "success": true, "session_name": "2026-09-16T20-00-<me>-OxLtF2",
  "entry_count": 628, "uploaded": true, "message": "session recovered via normal processing" }
```

Note the entry counts — **1151 and 628**, against the 991/447 recorded in the stale `meta.json`. The content was all on disk and complete. Nothing was missing; nothing was broken. The detector simply refused to look.

## Root cause

`internal/daemon/agentwork/session_finalize.go:426` — the draft guard, first statement in the per-session loop:

```go
if _, isDraft, metaErr := lfs.PreservedSessionIDAndDraft(sessionDir); isDraft || metaErr != nil {
    h.logger.Debug("skipping draft or unclassifiable session", "session", name, "err", metaErr)
    continue
}
```

Its own comment states the premise:

> Never touch a draft placeholder (ADR-029). It is an in-progress marker for a live recording, not an unfinished session awaiting anti-entropy: **it has no raw.jsonl to summarize and no artifacts to repair** […]

That premise is false for an abandoned draft. Mine had `draft: true` *and* 1151 entries of raw content. `draft: true` does not mean "live" — it means "was live when last written." The flag is set at start and cleared at finalize, so any session whose process dies in between is frozen with it set.

The check that could tell those two apart is 18 lines further down, at `:444-448`:

```go
// check for active recording
recPath := filepath.Join(sessionDir, recordingMarker)
if recInfo, statErr := os.Stat(recPath); statErr == nil {
    stale, recAge, method := isStaleRecording(recPath, recInfo, h.pidLookup, h.logger)
```

And `isStaleRecording` (`:1945`) already does precisely the right thing:

> if we have a PID, check liveness — **dead process = stale immediately**, live process = never stale

My PIDs were long dead. That branch would have fired on the first cycle. It never ran, because `continue` at `:426` executes first.

Worth noting the time-based fallback would not have saved this either: `staleRecordingThreshold = 24 * time.Hour` (`:48`), and these were 21.8h and 20.7h old. They were still three hours short of even the backstop — and the backstop is the wrong mechanism anyway when a definitive liveness signal is sitting right there.

The same ordering is duplicated at `:664-668`.

## Why this is worth its own fix

1. **It is silent.** `items=0` is indistinguishable from "everything is healthy." Nothing in `ox doctor`, `ox daemon status`, or the logs names the skipped sessions — the skip is a `Debug` line. A user has no way to learn their work is stranded.
2. **`--force-session-uploads` does not force.** The flag's entire purpose is to override normal gating for exactly this situation, and this gate is upstream of it.
3. **It has no time bound.** This is not "finalized late." A draft whose process died is skipped on *every* cycle forever. Mine would still be stranded today.
4. **It masks and outlives other fixes.** I hit this behind the ledger wedge in #956. After clearing that corruption the drafts were *still* stuck, which made a correct repair look like it had failed and sent me hunting for a second problem in the wrong place.

## Proposed fix

1. **Check liveness before honoring the draft flag.** If `.recording.json` names a PID and that PID is dead, the session is not a live draft — let it through to the existing recovery path. Narrowest correct change: move the staleness probe above the draft guard, and gate the guard on `isDraft && !stale`.
2. **Keep the `metaErr != nil` half fail-closed.** That arm protects against the #956 corruption class and should stay as-is — but it deserves its own branch and its own log line, since "draft" and "unparseable" are different conditions that currently share one skip and one message.
3. **Make the skip visible.** Promote skipped-with-content from `Debug` to a counter surfaced by `ox doctor` — e.g. `sessions skipped: 2 (draft, process dead)`. Any session skipped while holding a non-empty `raw.jsonl` is a strong anomaly signal and should never be silent.
4. **Make `--force-session-uploads` actually force.** Have it bypass the draft guard for sessions with a dead PID, since that is the one flag a user reaches for when they already believe something is stuck.

## Repro

1. Start a Claude Code session with ox recording; let it accumulate real turns.
2. Kill the client without a clean stop (force-quit, or close the terminal) so `meta.json` keeps `draft: true` and `.recording.json` survives.
3. Keep the daemon alive (other repos/sessions active).
4. Wait past a doctor interval, then run `ox doctor --force-session-uploads`.
5. `session finalize detect complete ... items=0`; `ox agent list` still shows `rec`; the session never publishes.
6. `ox agent <id> session recover` finalizes and uploads it immediately.

## Related

- #956 — the ledger wedge I was untangling when I hit this; the `metaErr != nil` half of the same guard is its fail-closed arm. Details in [this comment](https://github.com/sageox/ox/issues/956#issuecomment-5718131289).
- #962 — separate misclassification found in the same incident.
- #235 (closed) — related but distinct: that was orphans left when the **daemon dies** with the client. This is the inverse — the daemon is alive and healthy for 23h and still never detects them.

deepkawal:
Your coding session's transcript actually reaches the ledger after you close the window. Today, if the client process dies without a clean stop, that session is stranded on disk **permanently** — invisible to anti-entropy, invisible to `ox doctor`, and invisible to you.

I hit this on my own machine: two sessions holding **1151 and 628 entries** sat unuploaded for 21 hours behind a daemon that had been running and syncing the whole time. `ox doctor --force-session-uploads` looked at them twice and reported nothing to do.

## What was broken

`draft: true` means *"was live when last written"*, not *"is live"*. The flag is set at recording start and cleared only by finalize, so a client that dies in between freezes it set forever.

`detectInDir` honored that flag as its **first** statement — before ever checking whether the owning process still existed:

```mermaid
flowchart LR
    A["session dir"] --> B{"draft: true?"}
    B -->|yes| C["skip, Debug log"]
    B -->|no| D["liveness probe<br/>isStaleRecording"]
    D --> E["dead PID → reclaim ✓"]
    style C fill:#641220,stroke:#a4133c,color:#fff
```

`isStaleRecording` already implements exactly the right rule — *"dead process = stale immediately"* — and it was **unreachable for every draft**. The time-based backstop wouldn't have helped either: `staleRecordingThreshold` is 24h and these were 21.8h and 20.7h old.

Three things made this expensive to find:

- **It is silent.** `items=0` is indistinguishable from a healthy ledger. The skip was a `Debug` line.
- **It has no time bound.** Not "finalized late" — skipped every cycle, forever.
- **`--force-session-uploads` could not force it.** The gate sits upstream of the flag.

## What this ships

- **Liveness probe hoisted above the draft guard**, and keyed on the *evidence* rather than the bare verdict: `abandoned := hasRecordingMarker && recMethod == stalenessPIDDead`. The guard is now `isDraft && !abandoned`.
- **Age alone never overrides a live draft.** `isStaleRecording` also reports stale via `time_threshold`, which fires precisely when there was no PID to ask — unreadable or unparseable marker, or a legacy one predating `parent_pid`. That means *"nobody could be asked and 24h passed"*, never *"the owner died"*, and sessions here routinely run longer than 24h. Accepting it would clear a **live** draft's `.recording.json` and queue its transcript mid-write, dropping every entry written afterward. The 24h backstop is untouched for the pre-existing **non-draft** recovery path, where nothing is claiming the owner is alive.
- **A draft with no `.recording.json` stays skipped.** Absence of evidence is not proof of death, and that is precisely the shape of the ordinary git-tracked placeholder the CLI refreshes every turn (ADR-029's normal path is untouched).
- **The `metaErr != nil` arm split into its own fail-closed branch** with its own message. "Draft" and "unparseable meta.json" are unrelated conditions that were sharing one skip and one `Debug` line. Behavior is unchanged — an unreadable `meta.json` must still skip, never fall through to `recoverRawFromSessionFile`, which would write transcript bytes onto the git-tracked `raw.jsonl` and break LFS linkage for the whole team.
- **Skips are no longer silent.** A session skipped while holding a `raw.jsonl` with real entries is the strongest anomaly signal this scan produces. `detect complete` gains `skipped_with_content=N`, and a separate `Warn` names the sessions so you can act without guessing. The counter uses the same floor finalization does (`session.HasSubstantiveEntries`), so the header-only placeholder every session carries between start and first turn stays a routine skip instead of warning on a healthy ledger every cycle.
- **The same guard in `DetectOrphanedForAgent`** got the same treatment, gated on `ownerGone`.

**`--force-session-uploads` needed no separate bypass.** `ForceDetect` routes through `Detect` → `detectInDir`, so fixing the root gate fixed the flag. `TestForceDetect_ReachesAbandonedDraft` locks that in.

## What is deliberately NOT weakened

| Guarantee | Still holds because |
|---|---|
| Live draft never touched | `abandoned` requires a **dead-PID verdict** (`pid_dead`) — a live PID, a PID inside `GhostGracePeriod`, *no PID to ask* (`time_threshold`, incl. unreadable/unparseable/legacy markers), and no marker at all **all** still skip |
| LFS linkage | `sessionsDir != <ledger>/sessions` is unchanged and is what actually gates recovery writes; the draft guard only decides whether to *look* |
| Fail-closed on corrupt meta | Checked *before* any draft decision or recovery — the liveness probe above it only reads, so a dead PID does not buy a pass through it |
| `Detect` does not mutate | Draft flag and `ses_` id survive detection; `ClearDraft` remains finalize's job |

## Test Plan

`go build ./...`, `go vet`, `gofmt` clean. `golangci-lint run -c .config/golangci.yml ./internal/daemon/agentwork/` — **0 issues**. `go test ./internal/daemon/agentwork/` — **ok, 135.6s**. `go test ./cmd/ox/` — **ok, 558.6s**. `go test ./internal/session/... ./internal/daemon/` — ok (one unrelated flake in `internal/session/adapters`, `TestExternalAdapter_TimeoutKillsDescendants`, from CPU contention; green 3/3 in isolation).

6 new tests + reworked rows in the existing draft suite. Every row below was proven red-first — the production change reverted, the tests kept — then green after:

| Case | Asserts |
|---|---|
| `TestDetect_ReclaimsAbandonedDraft` (8 rows) | abandoned draft reclaimed in **both** ledger and cache trees; **live** draft still skipped; **no-marker** draft still skipped; **legacy marker (no `parent_pid`) past 24h** still skipped, marker intact; unreadable meta + dead PID still fails **CLOSED**; two negative controls — a dead-PID non-draft *and* a legacy-marker non-draft past 24h both still yield work |
| `TestDetectOrphanedForAgent_ReclaimsAbandonedDraft` | the second site, same invariants |
| `TestForceDetect_ReachesAbandonedDraft` | through the real `Manager` |
| `TestDetect_AbandonedDraftMetaSurvivesReclaim` | detection does not clear the draft flag or rotate the id |
| `TestDetect_WarnsAboutSessionsSkippedWithContent` / `_NoWarnWhenNothingIsStranded` | the new signal fires only when entries are actually stranded — neither the empty placeholder nor a header-only `raw.jsonl` counts |

**Two vacuous-pass traps, both real:** the pre-existing fixture helper only ever created a **dead**-PID marker. After this change that selects the *reclaim* path, so every existing "must skip" row would have passed **vacuously**. Split into `staleRecordingMarker` (dead), `liveRecordingMarker` (`os.Getpid()`) and `legacyRecordingMarker` (no `parent_pid` at all), with `heartbeatPID` set so the pre-existing cross-check can't short-circuit before the draft guard runs — the guard is now genuinely the thing under test. And the *"`ses_` id must not rotate"* assertion was comparing `""` to `""`: `writeFinalizedSession` never wrote an id, so it held with the whole preservation path deleted. The fixture now seeds one and both sides are pinned to it.

<details>
<summary>Checklist</summary>

- [x] Tests added
- [ ] CHANGELOG — N/A: daemon-internal detection fix, no CLI surface change
- [ ] Breaking change — N/A
- [x] `/security-review` — N/A: no `internal/auth/`, `internal/mcp/`, `raw_writer.go`, `prepush_scan.go`, `internal/upgrade/`, or `go.sum`

</details>

## Known follow-ups

- **`internal/doctor/session.go`'s `SessionOrphanedCheck.Run`** (the `ox doctor --fix` path) computes staleness independently via `status.StaleRecordings`, with no reference to `PreservedSessionIDAndDraft`. It may now be the last place with divergent draft-vs-stale semantics. Untouched here deliberately — separate issue.
- **No `ox doctor` stdout counter.** `--force-session-uploads` returns `BackgroundStarted` immediately and routes to `ox daemon logs`, so the count has no synchronous channel; surfacing it there would mean a new `DoctorResponse` field. The `Warn` lands in exactly the stream doctor points at.
- **`staleRecordingThreshold` left at 24h.** The definitive signal now fires first; changing the backstop is a separate call.

Fixes #966

SageOx-Session: https://sageox.ai/c/ses_01a0b039-02cc-74fd-81ad-beb272cd2e97

---
<a href="https://app.blacksmith.sh/sageox/codesmith/ox/pr/972"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792271647&installation_model_id=433550&pr_number=972&repository=sageox%2Fox&return_to=https%3A%2F%2Fgithub.com%2Fsageox%2Fox%2Fpull%2F972&signature=f34c6f56d6c3cdbc84aea00b815a71afd97513a95d755ac1f6f8cda479ca0fc3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Recovered abandoned draft sessions with recorded content after confirming their recording process is inactive.
  * Continued protecting active drafts and drafts without sufficient abandonment evidence.
  * Improved fail-closed handling for unreadable session metadata.
  * Added warnings for skipped sessions that may contain stranded content.
  * Excluded empty and header-only sessions from stranded-content warnings.
  * Ensured completed cached sessions remain queued for upload.
  * Preserved recovery behavior for eligible non-draft sessions.


deepkawal:
@greptileai

deepkawal:
@greptileai

Issue #767, opened 2026-08-14
LFS double-encode deadlocks team-context sync; daemon leaves unresolvable rebase state (v0.12.0/v0.13.0)


morphenius:
*[Claude Code · Opus 5 · local CLI · posting through @morphenius's account on behalf of the Softmax team]*

Filing this as an AI coworker that spent a session repairing a live alpha install — the findings below are mine, gathered first-hand from that repair, and posted with Michael's go-ahead. Private identifiers (team IDs, internal remote URLs, session links, local paths) are redacted; happy to supply specifics privately if any of it is hard to reproduce.

**Versions:** ox v0.12.0 (build 260719-194112) → v0.13.0 (build 260814-010901, commit `2bfaa5ca`, go1.25.8), darwin/arm64
**Environment:** macOS 24.6.0, git-lfs 3.7.1, team-context repo on a sparse checkout (~98% of tracked files present)

---

## TL;DR

A Git LFS **double-encode** in a team-context repo produced three files that git can never bring to a settled state. The sync daemon retried against this permanently-failing condition, accumulated stashes without bound, and eventually left the repo in a **structurally impossible rebase state** that neither `git rebase --continue` nor `git rebase --abort` could clear. Team-context sync was dead for roughly an hour and required manual git surgery to recover.

The corruption is the priority. The remaining findings are about why diagnosis took so long: the daemon's suggested fix could not work, `ox doctor` did not detect the problem at all, and a `--fix` hint pointed at a cause that wasn't there.

All observed on one machine in one session — I can't tell which findings are universal versus specific to this install.

---

### What is stored

Three files under `discussions/**/keyframes/` have a committed LFS pointer whose target object is *itself an LFS pointer* rather than image bytes.

```
$ git cat-file -p :discussions/<redacted>/keyframes/<redacted>.jpg
version https://git-lfs.github.com/spec/v1
oid sha256:12cd7e09c5c18a4f69e6d08b22051e24366a13af7ff643c5d2465e7eb08dc075
size 129

$ git show HEAD:discussions/<redacted>/keyframes/<redacted>.jpg
version https://git-lfs.github.com/spec/v1
oid sha256:12cd7e09c5c18a4f69e6d08b22051e24366a13af7ff643c5d2465e7eb08dc075
size 129
```

Resolving that oid yields a 129-byte payload that is the *inner* pointer:

```
version https://git-lfs.github.com/spec/v1
[...]

PR #974, opened 2026-09-17
fix(daemon): a failed conflict probe must never look like a sync

### LINKED ISSUE #962: Network timeout is reported as "unresolved conflicts" and suspends sync on a provably clean repo
## Summary

A **transport timeout suspends ledger and team-context sync and demands human intervention** — with a message telling the user their repo has merge conflicts it does not have. My laptop dropped DNS for a few minutes; when it came back, two of three repos stayed suspended, `ox doctor` reported `[CONFIRM REQUIRED]` conflicts on both, and clearing it required opening the internal ledger clone by hand to prove the conflicts weren't real.

**Root cause:** `ResolveAutostashConflicts` cannot distinguish "the index has conflicts" from "I failed to *find out* whether the index has conflicts." Any error from `git ls-files --unmerged` — including `context.DeadlineExceeded` — is returned to a caller that classifies every non-nil error as a merge conflict.

Observed on ox `0.15.0 (260912-023348)` / commit `05043a01`, macOS 26.2.

## The giveaway

```
[error] team_qdur30tb4b: team_qdur30tb4b has unresolved conflicts:
        inspect unmerged index: git ls-files: : context deadline exceeded
```

Note the empty field between `git ls-files:` and the error — that's the `%s` stderr slot in `RunGit`'s wrapper. The process produced no stderr because it was killed by context cancellation, not because git rejected anything. The repo was provably clean the whole time:

```
$ git status -sb          # => ## main...origin/main
$ git ls-files -u | wc -l # => 0
$ ls .git/ | grep -E 'MERGE|REBASE|CHERRY'  # => (nothing)
```

Yet: `Sync suspended after 5 consecutive failures`, and the issue persisted as `RequiresConfirm: true` long after connectivity returned.

## Root cause

`internal/gitutil/conflict.go:309-312` — every error from the probe is wrapped with a message asserting a conclusion the probe never reached:

```go
entries, err := listUnmergedEntries(ctx, repoPath)
if err != nil {
    return false, fmt.Errorf("inspect unmerged index: %w", err)
}
```

`listUnmergedEntries` (`internal/gitutil/rebase_resolve.go:416`) is a plain `RunGit(ctx, repoPath, "ls-files", "--unmerged")`, so it surfaces ctx cancellation as an ordinary command error.

`internal/daemon/sync_managed.go:291-303` — the caller then treats *any* non-nil `conflictErr` as a confirmed merge conflict:

```go
if conflictErr != nil {
    result.Err = errors.Join(result.Err, conflictErr)
    if result.Issue == nil {
        result.Issue = &DaemonIssue{
            Type:            IssueTypeMergeConflict,
            Severity:        SeverityError,
            Repo:            repoName,
            Summary:         fmt.Sprintf("%s has unresolved conflicts: %s", repoName, conflictErr),
            RequiresConfirm: true,
        }
    }
}
```

An infrastructural failure that carries **no information about index state** becomes a sticky, human-confirmation-gated `IssueTypeMergeConflict`.

## Why this matters beyond the wrong string

1. **It's the common case, not an edge case.** Laptop sleep, VPN flap, coffee-shop wifi, DNS blip — every ox user hits this. A timeout is the single most likely error from this probe, and it's the one misclassified.
2. **The failure is sticky and self-perpetuating.** `RequiresConfirm: true` means it won't clear itself; it survives the network recovering. Both my team contexts stayed suspended after connectivity was fine.
3. **It sends users into the internal clone.** The remedy on offer is manual git surgery under `~/.local/share/sageox/<endpoint>/{ledgers,teams}/*` — a path users should never open, and the same one #956 flags as a bad user-facing instruction.
4. **It masks real problems.** In my incident a genuine ledger wedge (a corrupt `meta.json`, see #956) was sitting behind two *false* conflict reports. Three errors, one real. The noise made the real one hard to find, and `ox doctor` couldn't separate them.

## Proposed fix

1. **Don't assert a conclusion the probe didn't reach.** Classify a probe *failure* separately from a probe *result*. At minimum, return a typed sentinel (`ErrConflictProbeFailed`) so callers can branch, and reword to "could not determine index state" when the probe itself failed.
2. **Treat `ctx.Err() != nil` as retryable, not as an issue.** A cancelled/deadline-exceeded probe should log at debug and let the next cycle retry — never mint a `RequiresConfirm` issue, never count toward the consecutive-failure suspension.
3. **Never let a transient error set `RequiresConfirm: true`.** Human-approval gating should be reserved for states a human actually has to adjudicate. A network timeout is not one.
4. **Re-probe before surfacing.** Before raising `IssueTypeMergeConflict`, confirm with a fresh, non-cancelled context that the index really is unmerged. Cheap, and it makes a false positive structurally impossible.

A one-line variant of (2) removes most of the pain:

```go
if conflictErr != nil && !errors.Is(conflictErr, context.DeadlineExceeded) && !errors.Is(conflictErr, context.Canceled) {
```

but the typed-sentinel version in (1) is the one that keeps the next transport error from re-creating the class.

## Repro

1. Let the daemon run with all three repos syncing.
2. Sever DNS/network for long enough that a sync cycle's context deadline expires mid-`ls-files` (laptop sleep works).
3. Restore connectivity.
4. `ox daemon status` → `has unresolved conflicts: ... context deadline exceeded`, `Sync suspended after N consecutive failures`, `[CONFIRM REQUIRED]`.
5. `cd` into the named clone → clean tree, zero unmerged entries, no in-progress operation.

deepkawal:
A few minutes of dead DNS stops costing you a suspended ledger and a false accusation. Today a laptop sleep, VPN flap, or coffee-shop wifi blip makes the daemon tell you your clone has **merge conflicts it does not have** — behind a human-confirmation gate that does not clear when connectivity returns.

The message that started this, on a repo with a provably clean index:

```
[error] team_qdur30tb4b: team_qdur30tb4b has unresolved conflicts:
        inspect unmerged index: git ls-files: : context deadline exceeded
```

Note the empty slot between `git ls-files:` and the error — that is the stderr `%s`. The process produced no stderr because it was **killed by context cancellation**, not because git rejected anything. Meanwhile: `Sync suspended after 5 consecutive failures`, `[CONFIRM REQUIRED]`, and a remedy of manual git surgery inside an internal clone path.

## What was broken

`ResolveAutostashConflicts` could not distinguish *"the index has conflicts"* from *"I failed to **find out** whether it does."* Every probe failure got a message asserting a conclusion the probe never reached, and `sync_managed.go` treated **any** non-nil error as a confirmed merge conflict with `RequiresConfirm: true`.

An infrastructural failure carrying **zero information about index state** became a sticky, human-gated `IssueTypeMergeConflict`.

## The regression this PR caught in itself

The first commit fixed that by tagging probe failures with a sentinel and dropping them. **Adversarial review caught that the cure was worse than the disease**, and the second commit fixes it.

The sentinel tagged *every* `ls-files --unmerged` failure, not just cancellation. The drop path returned a result with no `Err`, no `Issue`, and no `Skipped` — and in `doPull`, a zero-value result falls past both guards into the **success** path: `ClearSyncFailures`, `RecordPullSuccess`, `lastSync = time.Now()`.

For a transient timeout that is fine. For a **durable** failure it is catastrophic:

```mermaid
flowchart LR
    A["corrupt .git/index"] --> B["ls-files fails<br/>identically forever"]
    B --> C["tagged, dropped<br/>zero-value result"]
    C --> D["doPull: SUCCESS<br/>clears failures, advances lastSync"]
    D --> E["repo never syncs again<br/>every signal says healthy"]
    style E fill:#641220,stroke:#a4133c,color:#fff
```

`isValidGitRepo` only runs `rev-parse --git-dir` and never reads the index, so nothing upstream catches it. Three consecutive cycles over a corrupt index, pre-fix:

```
cycle 1: Skipped=false SkipReason="" Err=<nil> Issue=<nil>
cycle 2: Skipped=false SkipReason="" Err=<nil> Issue=<nil>
cycle 3: Skipped=false SkipReason="" Err=<nil> Issue=<nil>
```

**The first commit's own test asserted this as intended behavior**, on the premise "an unreadable index is the same class as the incident's timeout." It is not — a DNS blip self-heals; a truncated index does not. That test is reworked here, premise comment and all.

## What this ships

**The verdict now comes from a fact, not from an error's shape.** The three-way switch and `isUndeterminedIndexState` are deleted; `classifyAutostashFailure` **always** re-probes the index first, on `context.WithoutCancel(ctx)` so the cancellation that killed the original probe cannot kill the confirmation — and, after review round 2 below, up to `reprobeConfirmAttempts` (3) times before it will call a clone broken.

| Re-probe result | Verdict | Result shape |
|---|---|---|
| readable, unmerged entries | genuine conflict | `Err` + `IssueTypeMergeConflict`, `RequiresConfirm: true` — unchanged |
| readable, clean, pull never ran | transient | `Skipped` + `SkipReason: "unconfirmed conflict"` — no `Err`, no `Issue` |
| readable, clean, pull already ran | transient | untouched; the pull's own verdict stands |
| **unreadable** | **durable** | `Err` set, `IssueTypeRepoIntegrity`, `RequiresConfirm: **false**` |
| **abandoned** (round 4) | **nothing proved** | never a success: the cycle becomes `Skipped` + `SkipReason: "unconfirmed index read"`, whether or not a pull ran — no `Err`, no `Issue`, and it does **not** retire a standing integrity issue |

- **New `IssueTypeRepoIntegrity`.** All 15 existing types were checked; none conveys "git cannot read this clone." `merge_conflict` asserts a conflict was *found* — the exact false claim this PR removes.
- **`skipReasonUnconfirmedConflict` was verified against both consumers** before being chosen: `doPull` advances `lastSync` only for `"remote unchanged"` / `"recently fetched"`, and `sync_team.go` only special-cases `"rebase in progress"`. It takes the quiet no-op skip path, not a weaker form of the same bug.
- **`ClearIssue(IssueTypeRepoIntegrity, …)` on both success paths** (`sync.go`, `sync_team.go`). Without them the issue outlives the repair, the way `RebaseStuck` did before ox-baz5.3. Round 2 adds the same clear to the skips that also prove the index readable — see below.

What a durable failure looks like now — loud, and **not** `[CONFIRM REQUIRED]`:

```
[error] team_qdur30tb4b: git cannot read the index at /…/.git/index, so this repo
        is not syncing: could not determine index state: git ls-files --unmerged:
        fatal: .git/index: index file smaller than expected: exit status 128
```

## Review round 2: a durable verdict must still be retryable

Greptile caught this PR re-creating **the same failure class it exists to prevent**, one door over. `classifyAutostashFailure` declared the index unreadable from a **single** failed read — but a read fails identically whether the index is corrupt or a git got killed mid-run, a fork hit `EAGAIN`, or the filesystem stalled. The resulting error matches no substring in `isTransientSyncError`, so in `doTeamSync` it took the **fingerprint** branch: two cycles against the same dirty worktree set `SyncSuspended` and zeroed `NextSyncAttempt`. Permanent, unattended-unrecoverable, on a healthy machine — the [#906 network-blip incident](https://github.com/sageox/ox/issues/906) shape.

Reproduced before fixing: `attempt 3: an unreadable index must retry with backoff, never suspend permanently — Should be false`.

**Fix, per the repo's retry rule (enumerate what is settled; keep the unclassifiable retryable under a bounded attempt count):**

1. **Bounded confirmation.** `reprobeIndexConflicts` now takes 3 independent reads, 250ms apart, and reports unreadable only when **all** fail. A corrupt index fails all three in milliseconds, so the durable case is still loud on cycle 1 — the visibility guarantee the second commit added is untouched.
2. **Bounded backoff, never suspension.** `doTeamSync` routes a **pre-pull** `gitutil.ErrConflictProbeFailed` to `RecordSyncFailure` (capped exponential backoff) instead of `RecordSyncFailureFingerprint`. Round 2 tested the sentinel alone and justified it with "a probe failure aborts the cycle before any fetch" — **true only of the pre-pull probe**, and corrected in round 3 below. Durable stays visible as `IssueTypeRepoIntegrity` — which permanent silent suspension would only bury.

**Also from CodeRabbit:** a repaired clone kept a stale `repo_integrity` issue forever once its remote stopped changing, because every cycle dedups to a skip and skips never reach the clear-on-success path. `skipProvesIndexReadable` now clears it on the three skips that are only reachable **after** a successful index read (`remote unchanged`, `recently fetched`, `unconfirmed conflict`) and on no others — the lock and rebase skips return before anything reads the index, and round 3's `unconfirmed index read` is the case where the read never finished.

**A doc comment corrected with evidence, not assumption.** The original claimed the bare `context` checks caught cancellation deeper inside `ResolveAutostashConflicts`. A scratch program disproved it:

```
deadline during run     err="signal: killed"             isDeadline=false
deadline before start   err="context deadline exceeded"  isDeadline=true
```

`exec.CommandContext` returns the context error **only** when the context was already dead before `Start`. That is now the documented reason for not branching on error shape at all. `ErrConflictProbeFailed`'s own doc was also wrong — it told callers to branch on it, but an unreadable index carries it too; it now says so explicitly.

## Review round 3: a joined error is not a pre-pull failure, and a confirmation must not outlive a shutdown

Round 2's own fix was over-broad in one direction and under-bounded in another. Greptile filed both.

### 1. Joined errors bypassed the suspension that bounds autostash pile-up

`doTeamSync` routed on `errors.Is(err, gitutil.ErrConflictProbeFailed)` alone. But `classifyAutostashFailure` **joins** that sentinel onto an error a pull already produced, so a cycle where the pull **ran**, failed deterministically on a dirty worktree it can never settle, and *then* failed its post-pull index read carried the identical sentinel — and took ordinary backoff. That is exactly [#767](https://github.com/sageox/ox/issues/767): a pull retried forever, stacking an autostash entry every cycle, with the fingerprint guard skipped.

Round 2's report justified the sentinel check with "a probe failure aborts the cycle before any fetch, so #767 cannot reopen." **That is true only of the pre-pull probe.** The post-pull call site joins onto a pull that already executed.

**Fix — route on a structural fact, not on error identity.** `pullManagedRepo` already tracked `pullRan`; it is now a field on `ManagedRepoPullResult`, surfaced through `pullTeamContext` as `teamPullOutcome`, and the decision moved into one named, directly-testable predicate:

```go
func teamFailureTakesBackoff(err error, pullRan bool) bool {
	if isTransientSyncError(err) {
		return true
	}
	return !pullRan && errors.Is(err, gitutil.ErrConflictProbeFailed)
}
```

`pullRan` is a fact about the cycle that no joined error chain can report. The settled-vs-unclassifiable rule is unchanged: network failures and *pre-pull* probe failures stay retryable under bounded backoff; everything a pull did stays eligible for suspension.

### 2. A confirmation could hold daemon shutdown for ~45 seconds per clone

Every confirmation attempt ran under `context.WithoutCancel` with a 15s cap, so three failed reads plus their delays made one probe uninterruptible for **~45.5s**. Team sync waits on every bubble's confirmation *serially*, and daemon shutdown waits on the scheduler goroutine behind them.

Discarding cancellation for **one read** is the point — a dead parent context must not get to decide "durable" (#962). Discarding it for the **whole ladder** is a bug.

**Fix — the confirmation gets its own budget, collapsed on shutdown:**

| Bound | Value | What it protects |
|---|---|---|
| `reprobeAttemptTimeout` | 5s (was 15s) | one read; the granularity at which cancellation is ignored |
| `reprobeTotalBudget` | 20s | the whole ladder on a healthy daemon — larger than 3×5s + delays, so it never truncates a healthy run |
| `reprobeShutdownDrain` | 1s | all that remains once the sync context is seen canceled *between* two reads |

The drain, not an immediate return, is the load-bearing choice: **a corrupt index fails every read in microseconds**, so the fast, informative ladder still completes inside the drain and still reports `IssueTypeRepoIntegrity` on that very cycle. Only the slow, uninformative ladder is cut short — and a ladder cut short returns the new `errProbeAbandoned`, which `classifyAutostashFailure` reads as "nothing proved," never as a durable fault. Its pre-pull skip reason is the new `"unconfirmed index read"`, deliberately **excluded** from `skipProvesIndexReadable`: retiring a standing integrity issue needs a read that actually finished.

Three invariants now hold simultaneously, each with its own test: a durably corrupt index is visible on cycle 1 as `IssueTypeRepoIntegrity`; a plain cancellation never mints a `RequiresConfirm` merge conflict; and a deterministic post-pull failure still reaches suspension.

### 3. Changed-line coverage ratchet (CI `coverage` was red)

`88.2% (15/17)` against a 90% floor. Both uncovered statements were the `ErrConflictProbeFailed` tagging paths in `internal/gitutil/rebase_resolve.go` — exercised only from `internal/daemon`, whose coverage does not instrument another package. Closed with **real tests in `internal/gitutil`**, not an `EXCEPT`: `TestHasUnmergedEntries_UnreadableIndexIsATaggedFailure` and `TestResolveRebaseAcceptTheirs_UnreadableIndexIsATaggedFailure`, both over a corrupt `.git/index`.

## Review round 4: stop fixing instances, enforce the invariant

Greptile found the third variant of one bug, and that is the finding worth acting on. Each round shipped something **unproven** reported as success, and each fix was correct for its instance while leaving a narrower version of the class alive:

| Round | The unproven thing | How it read as success |
|---|---|---|
| 1 | every probe failure, dropped | zero-value result → `doPull` success path |
| 3 | joined post-pull probe error | routed to backoff, bypassing suspension |
| 4 | **abandoned confirmation after a successful pull** | the pull's verdict preserved, index state unknown |

Round 3's branch comment said the quiet part: *"no conflict was confirmed, so the pull's own verdict must not be overwritten — a successful pull really did sync."* With an abandoned confirmation that reasoning is false. **`git pull --rebase --autostash` exits zero after a rebase whose autostash re-apply left unmerged entries behind** — that is the entire reason this cycle re-reads the index — so with no completed read a broken worktree and a clean one are the same bytes. The successful verdict flowed to `ClearSyncFailures`, `ClearIssue(repo_integrity)`, `RecordPullSuccess`, `lastSync`.

**Fix — one invariant, one place, instead of a fourth special case:**

> A result may leave `classifyAutostashFailure` readable as a completed sync **only** when a read that ran to completion proved the index clean.

Three structural changes carry it:

1. **`indexProof` replaces the `(conflicted bool, probeErr error)` pair** as the only thing the classification may branch on. A bool-and-error pair invites `err != nil ⇒ broken` and `!conflicted ⇒ clean`; the enum makes `proofUnknown` — *nothing was established* — a value you must handle, not a gap between two others.
2. **`applyIndexProof` is pure** (`classifyAutostashFailure` = probe + classify + apply), so the decision is testable over its whole input space without staging real git index states.
3. **One guarded exit.** Every branch funnels through `withheldUnprovenSuccess`, which downgrades any success-looking result to the skip whenever the proof is not `proofClean`. It is written as a blanket "not clean" rather than "== unknown" **so the next branch that forgets to fail the cycle trips it instead of shipping.**

The remedy is a **skip, not an error**, deliberately: nothing is known to be wrong, so the cycle stays ordinarily retryable — an error would feed the consecutive-failure backoff and, with `PullRan` set, `teamFailureTakesBackoff`'s permanent fingerprint suspension, turning every laptop-lid-close into a repo that stops syncing. `skipReasonUnconfirmedIndex` is the reason precisely because `skipProvesIndexReadable` says false for it: a read that never finished proves nothing in **either** direction, so it must not retire a standing `repo_integrity` issue either.

All four constraints hold simultaneously, each with its own test: (a) a durably corrupt index is visible on cycle 1 as `IssueTypeRepoIntegrity`, `RequiresConfirm: false`; (b) plain cancellation never mints a `RequiresConfirm` merge conflict; (c) a deterministic post-pull failure still reaches fingerprint suspension; (d) an abandoned confirmation is never a sync, stays retryable, mints nothing and retires nothing. A genuinely clean, **completed** confirmation after a successful pull still records success.

## Test Plan

`go build ./...`, `go vet ./internal/gitutil/ ./internal/daemon/...`, `gofmt -l` clean. `golangci-lint run -c .config/golangci.yml ./internal/gitutil/... ./internal/daemon/...` — **0 issues**. `go test` — `internal/daemon` **261.9s**, `internal/gitutil` **48.6s**, `internal/doctor` **2.3s**, all ok (round-4 re-run; `cmd/ox` **427.5s** ok as of an earlier commit, untouched since). Full tier `make test-all`: 21,898 tests, one unrelated flake in `internal/codedb/index` (`TestGCDirtyIndexes/removes_stale_overlay`, `signal: segmentation fault` from a child process under the full fan-out; **passes in isolation**, different package, no dependency on the changed code).

Coverage ratchet, computed exactly the way CI does (`coverage_ratchet.py` with and without `--diff-base` against the merge base `597ac24`): protected package floors all **PASS** (`internal/daemon` 66.6% vs floor 63.5%), changed-line **100% (44/44)** vs a 90% floor. No `EXCEPT` entry was added. Because a pre-existing dated `EXCEPT` from PR #868 covers `sync_managed.go` at file level, the new statements were additionally verified directly — `go tool cover -func`: `classifyAutostashFailure`, `classifyIndexProof`, `applyIndexProof`, `withheldUnprovenSuccess`, `resultReadsAsSuccess`, `provesIndexState`, `String` — **100.0% each**.

Red-before-green reproduced against the first commit, matching the report above exactly, then passing.

| Test | Covers |
|---|---|
| `…_UnreadableIndexIsVisibleAndNotAConflict` | reworked: keeps "not a `RequiresConfirm` merge conflict", **adds** `Err` non-nil + `IssueTypeRepoIntegrity` + summary names the index path |
| `…_CorruptIndexNeverLooksLikeSuccess` | 3 cycles, asserts on doPull's literal predicate `!Skipped && Err == nil` |
| `TestClassifyAutostashFailure` | 6 cases, every branch, each under a **pre-cancelled** parent context |
| `TestReprobeIndexConflicts` | 4 subtests — **first coverage anywhere** of `HasUnmergedEntries`, which had zero test references |
| `…_GenuineUnmergedIndexStillMintsConflictIssue` | unchanged and still passing — real conflicts are not suppressed |
| `TestDoTeamSync_UnreadableIndexDoesNotSuspend` | **round 2, red-first**: 4 cycles through the real `doTeamSync` classification over a **dirty** worktree; suspends on attempt 3 without the fix |
| `TestReprobeIndexConflicts_BoundedConfirmation` | **round 2, red-first**: one failed read is retried, not called durable; every read failing is still durable |
| `TestPullTeamContext_ReadableSkipClearsStaleIntegrityIssue` + `TestSkipProvesIndexReadable` | **round 2, red-first**: the stale-issue clear, and which skips may trigger it |
| `TestTeamFailureTakesBackoff` | **round 3, red-first**: 6 rows; the joined post-pull error must SUSPEND, the pre-pull one must back off |
| `TestDoTeamSync_PostPullFailureStillSuspends` | **round 3, red-first**: end-to-end through the real `doTeamSync` — pull runs, fails on a dirty worktree, post-pull probe fails, suspension still lands |
| `TestReprobeIndexConflicts_ShutdownStopsASlowLadder` | **round 3, red-first**: a slow ladder is cut short on shutdown and returns `errProbeAbandoned` |
| `TestReprobeIndexConflicts_ShutdownStillConfirmsAFastLadder` | **round 3**: a corrupt index is still called durable during shutdown — the drain must not hide it |
| `TestClassifyAutostashFailure_AbandonedConfirmationProvesNothing` | **round 3 + round 4, red-first**: abandoned never becomes `IssueTypeRepoIntegrity` and never overwrites a failed pull's verdict — plus the new round-4 case, a **successful** pull with an abandoned confirmation, red on `29ee15bd` with `Skipped=false SkipReason="" Err=<nil>` |
| `TestApplyIndexProof_NoUnprovenResultLooksLikeASync` | **round 4**: the full cross-product — 6 probe outcomes × `pullRan` × 4 incoming result shapes = **48 rows** — asserting `resultReadsAsSuccess` is true in exactly the one legal cell, plus that no unproven cycle claims a readable index, mints a `RequiresConfirm` issue, or adds an error of its own |
| `TestClassifyIndexProof` | **round 4**: the `(conflicted, probeErr) → indexProof` mapping, including that an abandoned ladder outranks the read error it still carries in its chain |
| `TestHasUnmergedEntries_…` + `TestResolveRebaseAcceptTheirs_UnreadableIndexIsATaggedFailure` | **round 3**: the two uncovered statements the coverage ratchet failed on |
| `TestSyncScheduler_PullTeamContext_CorruptRepoRenameFailure` | **round 3**: a failed move-aside is never reported as a sync (also the last uncovered changed statement) |

Independently verified outside the PR's own suite, both directions:

```
durable corrupt index:   Skipped=false Err=true Issue=true looksHealthy=false   ×3
transient cancellation:  Skipped=true SkipReason="unconfirmed conflict" Err=false Issue=false
```

<details>
<summary>Checklist</summary>

- [x] Tests added
- [ ] CHANGELOG — N/A: daemon-internal classification fix, no CLI surface change
- [ ] Breaking change — N/A
- [x] `/security-review` — N/A: no `internal/auth/`, `internal/mcp/`, `raw_writer.go`, `prepush_scan.go`, `internal/upgrade/`, or `go.sum`

</details>

## Known follow-ups

- **`isValidGitRepo` still only runs `rev-parse --git-dir`.** Making it read the index would cause `ValidateIntegrity` to classify corruption as `CorruptRepo` — which for team-context and bubbles means *move the clone aside and re-clone*: destructive, potentially dropping unpushed work, Sacred-tier-adjacent. Deliberately a separate decision.
- **No `ox doctor` check for `repo_integrity`** yet; it surfaces via the generic `FormatLine` renderer. A dedicated check with a `--fix` re-clone is the natural next step.
- **`sync_managed.go`'s second conflict-minting site** matches `strings.Contains(statusOutput, "UU")` unanchored, so a filename like `UUID.txt` could trip a false merge conflict. Pre-existing, untouched, worth its own issue.

Touches `internal/gitutil/conflict.go`, which #973 also modifies — different regions of the same function, so expect at most a trivial append-vs-append conflict in `conflict_test.go` for whichever lands second.

Fixes #962

SageOx-Session: https://sageox.ai/c/ses_01a0b039-02cc-74fd-81ad-beb272cd2e97

---
<a href="https://app.blacksmith.sh/sageox/codesmith/ox/pr/974"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792271746&installation_model_id=433550&pr_number=974&repository=sageox%2Fox&return_to=https%3A%2F%2Fgithub.com%2Fsageox%2Fox%2Fpull%2F974&signature=ca6df7d6a8e01c3241baec5a5931fa7cdbb26456c7ab3c329ef3dec66b3d2691"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization error reporting when Git repository state cannot be inspected.
  * Unreadable repository state is now reported as a repository integrity issue instead of a confirmed merge conflict.
  * Genuine merge conflicts continue to require confirmation before resolution.
  * Integrity issues clear after successful synchronization or a readable skipped sync.
  * Incomplete checks preserve accurate sync outcomes and retry appropriately.
  * Team synchronization now distinguishes transient inspection failures from confirmed post-sync failures.


deepkawal:
@greptileai

deepkawal:
@greptileai

deepkawal:
@greptileai

deepkawal:
@greptileai

deepkawal:
@greptileai

deepkawal:
@greptileai

PR #1003, opened 2026-09-19
fix(ledger): verify pointer-shaped files by content in read sync

### LINKED ISSUE #1000: nested_stub rejects correctly hydrated files whose bytes look like an LFS pointer
Parent: https://github.com/sageox/sageox-monorepo/issues/4180 (private coordination). Implementation contract: #891. Reader contract: `docs/specs/ledger-read-sync.md`.

## What happens

`readFiles` declares any pointer-shaped worktree file that differs from HEAD's pointer blob a `nested_stub`, and returns before checking whether its bytes are the object HEAD commits:

```go
// internal/ledger/read_sync.go — readFiles
if gitOID != file.oid {
    if len(file.pointer) != 0 {
        // The worktree file is itself a pointer, and not the one HEAD
        // commits. It names some other object, so it is not this file's
        // content and no OID here would be the one worth reporting.
        return nil, missingHydration(ReadFailureDetail{Reason: "nested_stub", Path: name})
    }
    // ...reads HEAD's committed pointer, and requires
    //    size == pointerSize && lfsOID == committed oid
```

The branch right below it is already the correct LFS integrity check: parse the committed pointer, require the worktree bytes' SHA-256 and size to match it. The early return skips that check for exactly the files where it matters.

The comment's premise — "it names some other object, so it is not this file's content" — holds for a **stale** pointer left in the worktree. It is false when the committed object's content is itself pointer-shaped, and that is a real, recurring git-lfs corruption: an already-cleaned file cleaned again, so the pointer is stored as the object.

## Observed

A cold read sync of a real ledger (~27,200 session files) against the live read transport. Transport was clean — every object downloaded, `skipped` empty. The sync still ended `ready: false, error_class: missing_hydration, reason: nested_stub` on two files from a single session (`sessions/<session>/context-trace.jsonl` and `.../raw.jsonl`):

| | |
|---|---|
| Pointer HEAD commits | `oid sha256:9d530f74…`, `size 126` |
| Bytes on disk after hydration | 126 bytes: a pointer to `sha256:e3b0c442…`, `size 0` |
| `sha256(bytes on disk)` | `9d530f74…` — **equal to the committed oid** |

Both files are correctly hydrated by LFS's own definition: their bytes are the committed object, byte for byte. `git lfs smudge` would write the same file. `readFiles` rejects them anyway.

**The corrupt object has a universal fingerprint.** `9d530f74b243d7e8f27437991152e4a2f4a581b0fb6873c115aec2ee54aa5867` is the SHA-256 of the 126-byte pointer to the empty object (`e3b0c442…`, the SHA-256 of empty content). Every double-cleaned *empty* file in any repository stores this identical object, so this is not specific to one ledger.

## Why this matters

It is **permanent and deterministic**. The two files are two of ~27,200, yet no attempt can ever produce a ready checkout, and the resumable stage (#951) cannot help: a resume re-runs `readFiles` over the already-hydrated files and fails the same way, immediately. One double-cleaned file anywhere in a ledger's history makes it unreadable through this path forever.

## Required outcome

Decide `nested_stub` by content, not by shape. A pointer-shaped worktree file whose bytes match the committed pointer's oid and size is **hydrated**, and verification continues. Only a pointer-shaped file whose bytes do *not* match — a genuinely stale pointer — is `nested_stub`.

This needs no new logic: run the existing committed-pointer comparison before the early return, and keep `nested_stub` as the diagnostic when that comparison fails on a pointer-shaped file. It tightens the check rather than relaxing it — content hash instead of content shape.

## Acceptance

- A file whose committed object's content is a pointer (use `9d530f74…` / the pointer-to-empty object as the fixture) verifies as hydrated; the checkout reaches `ready: true`.
- A pointer-shaped worktree file whose bytes do not hash to the committed oid is still reported `nested_stub`.
- A resumed stage containing such a file resumes and verifies instead of failing on entry.
- The existing stale-pointer and `malformed_pointer` cases keep their current results.

Related, not in scope here: when verification fails this way, `coverage` and `hydration` counts read zero even though the stage is ~all hydrated — #988 fixed that for hydration-time failures, not for verification-time ones.

shrimalmadhur:
> guided&nbsp;by&nbsp;<picture><source media="(prefers-color-scheme: dark)" srcset="https://sageox.ai/sageox-wordmark-dark.png">SageOx</picture>&nbsp;&middot;&nbsp;<a href="https://sageox.ai/c/ses_01a0b6c5-7333-738a-9a3a-8ce80393ca85">Session</a>

Closes #1000.

## What broke

`readFiles` decided two failures by the **shape** of a worktree file before checking its **content**, and a third defect hid how far a failed sync had got.

| # | Defect | Effect |
|---|---|---|
| 1 | A pointer-shaped file that differs from HEAD's blob is `nested_stub` — returned before the committed-pointer comparison right below it ran (#1000) | An object whose own content is a pointer can never verify. No attempt reaches `ready`. |
| 2 | A file that *looks* like a pointer but does not parse is `malformed_pointer` — also returned before that comparison | Same, for an object whose content is a pointer this reader cannot parse, e.g. one naming an object above `OX_LFS_MAX_OBJECT_SIZE` |
| 3 | `verifyReadCheckout` replaces the result; when it stops before counting, the counts hydration took are discarded | A sync that landed ~every object reports `coverage.files: 0`, hydration `unknown` — reads like one that did nothing |

- **Observed (#1000).** A cold sync of ~27,200 session files over a clean transport (`skipped` empty) ended `ready: false`, `nested_stub`, on two files of one session. Their bytes hash to the committed oid `9d530f74…` — the 126-byte pointer to the empty object, which is what an empty file cleaned a second time stores, identical in every repository that has this corruption.
- **Permanent, and expensive.** The next attempt's `resumableReadStage` walks the stage with the same `readFiles`, rejects it, and re-clones — so every retry transferred every object again, then failed verification the same way.
- **Where the objects came from.** Until #934, `ParsePointer` rejected `size 0`, so session upload did not recognise the pointer to an empty file as a pointer and uploaded its text as content. #934 closed that path; ledgers written before it keep the objects in history. #1004 stops every upload path from storing a pointer as content.

## What this PR ships

```mermaid
flowchart TD
    W["worktree file"] --> E{"same bytes as<br/>HEAD's blob?"}
    E -- yes --> P{"what HEAD commits"}
    P -- "a pointer" --> STUB["stub: hydrate it"]
    P -- "an unparseable pointer" --> MP["malformed_pointer<br/>(unchanged)"]
    P -- "anything else" --> PLAIN["plain file"]
    E -- no --> C{"bytes are the object<br/>HEAD's pointer names?<br/>SHA-256 and size"}
    C -- yes --> H["hydrated<br/>whatever its shape"]
    C -- no --> S{"shape of the bytes"}
    S -- "a pointer" --> N["nested_stub"]
    S -- "an unparseable pointer" --> MP2["malformed_pointer"]
    S -- "anything else" --> D["dirty"]
    style H fill:#2a4a2a,stroke:#6c6,color:#fff
```

- **Content decides; shape only names the failure** (commits 1–2). The existing committed-pointer comparison now runs first for every worktree file that differs from HEAD's blob. Bytes that are the committed object are hydrated, pointer-shaped or not. When the comparison fails, the shape picks the reason exactly as before: `nested_stub`, `malformed_pointer`, or `dirty`.
- **Tightens, never relaxes.** A pointer-shaped file used to fail on shape; to pass now it must hash to the committed OID at the committed size. A pointer HEAD itself commits and this reader cannot parse is still `malformed_pointer` — never served as a plain file.
- **Every `readFiles` caller inherits it** — the pre-fetch inspection of a published checkout, `resumableReadStage`, `dehydrateReadFiles` (retains the object under `.sageox/cache/read-sync/objects/<oid>` by its committed OID), `hydrateReadFiles` (skips it as hydrated), and `verifyReadCheckout`.
- **Counts survive a verification that stops short** (commit 3). When verification returns before it counted — HEAD, its history, or a file fails it, or the budget runs out while it hashes the stage — the result keeps the counts hydration took. When verification counts, its counts win, as before.

| Verification | `coverage` / `hydration` before | after |
|---|---|---|
| stops before counting | `files: 0`, `unknown 0/0` | the counts hydration took |
| counts | its own | its own (unchanged) |

- **Contract.** `docs/specs/ledger-read-sync.md`: the two reason rows, one paragraph saying both are decided by content, and one sentence on counts when verification stops short. No schema or reason-vocabulary change.

## Test Plan

**Red-first**, hermetic fixture: real Git smart-HTTP server, real LFS batch and download routes. The #1000 object is the fixture; the test asserts its OID is `9d530f74…`.

| Break | Red |
|---|---|
| Unfixed `nested_stub` | `TestReadSyncLFSPointerShapedObjectVerifiesByContent/pointer_to_the_empty_object`: cold sync `Ready:false`, `nested_stub` on `sessions/doubled/raw.jsonl`. `TestReadSyncColdStageHoldingAPointerShapedObjectResumes`: resumed attempt `Ready:false`, `nested_stub`, object transferred **2** times — the stage was re-cloned, not resumed |
| Unfixed `malformed_pointer` | `…VerifiesByContent/pointer_this_reader_cannot_parse`: cold sync `Ready:false`, `malformed_pointer` |
| Unfixed counts | `TestReadSyncColdVerificationFailureKeepsTheHydrationCounts`: hydration `{unknown 0 0}`, expected `{complete 1 1}` |
| Accept by size alone (drop the SHA-256 comparison) | `…VerifiesByContent`: a same-size stale pointer reads ready |
| Report every mismatch as `dirty` | `…VerifiesByContent` and `TestReadSyncRejectsDamagedTrackedContent/different_stub`: `dirty`, expected `missing_hydration` |
| Drop the malformed diagnostic on a mismatch | `TestReadSyncRejectsDamagedTrackedContent/malformed_stub` and `…/pointer_this_reader_cannot_parse`: `dirty`, expected `missing_hydration` |
| Drop the refusal of an unparseable pointer HEAD commits | `TestReadSyncLFSUnparseableCommittedPointerIsNeverServed`: the pointer's text is served as a ready file |
| Always keep hydration's counts | `TestReadSyncNativeLifecycle`: a denied refresh (hydration never ran) reports `unknown`, expected verification's `complete` |

Fix restored: all green.

New/extended coverage:

- `TestReadSyncLFSPointerShapedObjectVerifiesByContent` (new file `read_sync_verify_test.go`) — for a parseable and an unparseable inner pointer: cold sync `ready` with hydration `{complete 1 1}` and the file byte-for-byte the object; warm refresh ready with no second transfer; `--check` ready; a same-size stale stub reports its reason with its path and is left untouched.
- `TestReadSyncLFSUnparseableCommittedPointerIsNeverServed` — HEAD commits a pointer this reader cannot parse: `malformed_pointer`, and no object is requested.
- `TestReadSyncColdStageHoldingAPointerShapedObjectResumes` — a stage holding the #1000 object is resumed (object transferred once) and publishes ready.
- `TestReadSyncColdVerificationFailureKeepsTheHydrationCounts` (`read_sync_stage_test.go`) — a file written into the stage while objects transfer fails verification; the result keeps hydration's counts, and they equal what `verifyReadCheckout` finds once the file is gone.
- `TestReadSyncRejectsDamagedTrackedContent` — the existing stale- and malformed-stub cases now pin `error_detail`, not only `error_class`.
- `TestReadSyncNativeLifecycle` — a denied refresh reports the counts verification took.

Gates:

- `go test ./internal/ledger/` — full package, real Git — passes
- `go test -race` on the new tests — passes
- `go test ./cmd/ox/ -run 'TestReadSync|TestHostedLedger|TestLedgerRead'` — passes
- `make lint` — 0 issues

## What readers see

- **Readers after the fix.** A session whose files are the #1000 object now reads as a stub to the guarded readers: `session.ClassifyRawFile` returns `RawPointerStub` for its `raw.jsonl`, and glance skips it (`internal/glance/harvest.go`). Its original file was empty, so nothing is lost.

SageOx-Session: https://sageox.ai/c/ses_01a0b6c5-7333-738a-9a3a-8ce80393ca85


shrimalmadhur:
@greptileai

shrimalmadhur:
@coderabbitai review

PR #1005, opened 2026-09-19
fix(cli): send JSON diagnostics to stderr

shrimalmadhur:
## What broke

- **Trigger:** A command calls `cli.PrintWarning` or `cli.PrintError` while JSON mode is active.
- **Before:** The diagnostic goes to stdout. If the command also emits a result, stdout contains two JSON documents and fails single-document parsing. Text mode already sends these diagnostics to stderr.
- **Example:** `ox uninstall --dry-run --json` in an uninitialized repository emits its diagnostic JSON on stdout despite failing.

## What this PR ships

- **Fix:** Route both helpers through the existing `PrintWarningTo` / `PrintErrorTo` implementations with stderr as the writer, removing the duplicate JSON branches.
- **Output contract:** Diagnostic JSON retains its `status` and `message` fields. Command results and command-specific response envelopes keep their stdout contract. Consumers of helper-generated diagnostics should read stderr.
- **Regression coverage:** Check both streams for warnings and errors in text and JSON modes; exercise the actual uninstall failure handler.
- **Documentation:** Record the stream contract in the CLI design system, consistent with [clig.dev](https://clig.dev/#the-basics).
- **Hook test reliability:** Wait for subprocess completion through the existing `runner.Wait()`. The fixed 300 ms delay raced the subprocess on a busy CI runner, causing a missing-output-file failure.
- **Git fixture isolation:** Pass the commit-signing override through `git -c` in the fixture helpers so they preserve inherited maintenance settings. Replacing `GIT_CONFIG_COUNT` had dropped the test runner's protection against background Git processes racing temporary-directory cleanup.

## Test Plan

- **Red/green proof:** Both JSON diagnostic cases and the uninstall regression fail before the fix and pass after it.
- **Compiled CLI checks:** Five probes cover `--json`, `OX_JSON=1`, `--json=false`, quiet-mode errors, and a warning followed by a JSON result.
- **Fast gate:** Final `make test` — 19,560 passed, 1,239 skipped; final lint clean.
- **Full gate:** `make test-preflight` — 22,469 full-suite tests passed (28 skipped), 10,014 slow tests passed (15 skipped), lint clean; run before the test-only CI reliability follow-ups.
- **Hook follow-up:** The previously failing hook test passed 100 runs under the race detector; final lint and fast tests also passed.
- **Git follow-up:** A regression proves both fixture helpers preserve inherited configuration (red before the fix, green after). That test and the sequential-conflict scenario passed ten runs under the race detector; the full Git utility suite also passed with race detection.
- **Formatting:** `make format` and `git diff --check`.

SageOx-Session: https://sageox.ai/c/ses_01a0b6cf-20e4-7e96-90dd-bdc8d83a0a79


shrimalmadhur:
@coderabbitai full review

shrimalmadhur:
@greptileai

